PR #24, opened 2015-10-01
wip observations dashboard

jshawl:


app/controllers/observations_controller.rb
```diff
@@ -1,3 +1,19 @@
 class ObservationsController < ApplicationController
-
+  def index
+    @group = Group.find(params[:group_id])
+    @student = Membership.find(params[:student_id])
+    @observation = Observation.new
```
mattscilipoti: Expected an index of observations to have @observations.  Can that be the interface that the form uses?

app/views/observations/index.html.erb
```diff
@@ -0,0 +1,19 @@
+<a href="../">&larr;</a>
+<h2><%= @student.user.username %></h2>
+
+
+<h3>Observations</h3>
+
+<%= form_for [@group, @student, @observation] do |f| %>
```
mattscilipoti: Would like to avoid 3 layers of nesting.  Does an observation for a student really need the group?  Or is it just an observation for a student?

PR #75, opened 2015-10-13
assignment: Default due_date

mattscilipoti:
Added specs for setting default Assignment#due_date
Using pry-byebug to support 'debugger' OR 'binding.pry'


adambray:
:+1:

PR #76, opened 2015-10-13
add view for names to get pictures of all students

andrewsunglaekim:


jshawl:
This build is failing because you need to update your branch with work on upstream.

Look at the output of `git diff --stat upstream/master` while on this branch

PR #85, opened 2015-10-14
autocreation of memberships

RobertAKARobin:
When a user is made member of a group, they are also made a member of all of that group's ancestors. When a user is made an *admin* member of a group, they are additionally made an admin of all of that group's *descendants*. Destroying a membership destroys all of its sub-memberships.


app/controllers/application_controller.rb
```diff
@@ -4,7 +4,7 @@ class ApplicationController < ActionController::Base
   # protect_from_forgery with: :exception
   before_action :authenticate
   helper_method :current_user, :current_user_lean, :signed_in?
-  rescue_from StandardError, with: :global_rescuer
+  # rescue_from StandardError, ActionController::RedirectBackError, with: :global_rescuer
```
adambray: why is this commented out?

app/controllers/users_controller.rb
```diff
@@ -3,7 +3,10 @@ class UsersController < ApplicationController
   skip_before_action :authenticate, except: [:show]
 
   def index
+    @is_su = is_su?
```
adambray: can we just make this a helper method so views can use it without having to add code to the controller?

app/models/group.rb
```diff
@@ -13,25 +13,25 @@ class Group < Tree
...
-      if sibling && sibling.id != self.id
-        errors.add(:title, "must be unique among this group's siblings.")
-      end
+  def has_parent
+    if self.class.all.count > 1 && !self.parent
+      raise "Each group has to have a parent group."
+    end
+  end
+
+  def has_unique_title_among_siblings
```
adambray: I'd rename this to `validate_uniqueness_of_title_among_siblings`

app/models/group.rb
```diff
@@ -13,25 +13,25 @@ class Group < Tree
...
-    if !self.parent
-      if self.class.all.count > 0
-        errors[:base].push("Each Group has to have a parent group.")
-      end
-    else
-      sibling = self.parent.children.find_by(title: self.title)
-      if sibling && sibling.id != self.id
-        errors.add(:title, "must be unique among this group's siblings.")
-      end
+  def has_parent
```
adambray: I'd rename this `validates_presence_of_parent`

adambray:
Made some suggestions, not really critical. Otherwise, looks good to me. :+1:

app/controllers/application_controller.rb
RobertAKARobin: I added it in the previous commit, and then commented it out because I wanted to actually see all the errors I was making.

RobertAKARobin: I feel like it's probably bad practice to have a global error handler.

PR #87, opened 2015-10-14
fix case where users don't have names

adambray:


adambray:
Merging this in now as it fixes a major bug in production (seed data didn't include this case, so I didn't detect it until just now)

PR #94, opened 2015-10-15
Show names in attendance, return username if no name is set

adambray:


RobertAKARobin:
Will this affect setting names? As in `@user.name = "joe"`.

PR #95, opened 2015-10-15
broke groups show page into mutiple pages bc takes forever to load

RobertAKARobin:


RobertAKARobin:
Also made user avatar cool and zoomable, and made the "Submit Attendance" button red and more prominent so Adam remembers it's there.

PR #113, opened 2015-10-16
added orphans path

RobertAKARobin:
For users with no memberships.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,7 +4,7 @@ class ApplicationController < ActionController::Base
   # protect_from_forgery with: :exception
   before_action :authenticate
   helper_method :current_user, :signed_in?, :is_garoot?
-  rescue_from StandardError, ActionController::RedirectBackError, with: :global_rescuer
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -4,7 +4,7 @@ class ApplicationController < ActionController::Base
   # protect_from_forgery with: :exception
   before_action :authenticate
   helper_method :current_user, :signed_in?, :is_garoot?
-  rescue_from StandardError, ActionController::RedirectBackError, with: :global_rescuer
+  # rescue_from StandardError, ActionController::RedirectBackError, with: :global_rescuer
```
jshawl: ?

app/controllers/users_controller.rb
```diff
@@ -2,6 +2,11 @@ class UsersController < ApplicationController
 
   skip_before_action :authenticate, except: [:show]
 
+  def orphans
```
jshawl: what is the purpose of orphans? what do they do?

jshawl:
otherwise :+1:

PR #121, opened 2015-10-19
updated refresh_memberships and moved user info to bottom of profile page

RobertAKARobin:


nolds9:
:+1:

app/models/group.rb
```diff
@@ -80,4 +80,12 @@ def bulk_create_memberships array, is_admin
     end
   end
 
+  def has_member? user
+    return self.memberships.where(user_id: user.id).count > 0
+  end
```
mattscilipoti: exists wold work well here: http://apidock.com/rails/ActiveRecord/Base/exists%3F/class

app/models/membership.rb
```diff
@@ -86,4 +78,15 @@ def has_descendants?
...
   end
 
+  def update_ancestor_memberships
+    user = self.user
+    checked_so_far = []
+    self.group.ancestors.each do |group|
+      next if checked_so_far.include?(group.id)
+      checked_so_far.push(group.id)
+      next if group.has_member?(user)
+      group.memberships.create!(user_id: user.id)
```
mattscilipoti: Recommend using object/model when possible: `group.memberships.create!(user: user)`

app/models/group.rb
RobertAKARobin: Oh, wow. Yes, it sure would.

app/models/membership.rb
RobertAKARobin: I forgot you can do that.

app/views/attendances/_counts.html.erb
```diff
@@ -0,0 +1,3 @@
+<% [nil, 0, 1, 2].each do |s| %>
+<span><%= Attendance.status_english(s) %>:<%=attendances.select{|a| a.status === s}.count %></span>
```
mattscilipoti: I expected a scope here `attendances.with_status(s)`

mattscilipoti: I recommend using `size`. http://web.archive.org/web/20100210204319/http://blog.hasmanythrough.com/2008/2/27/count-length-size

RobertAKARobin: I think `attendances` here is not necessarily an ActiveRecord collection; it may also be an array. That's a problem in and of itself -- I wasn't sure how to do that. See `#get_descendants` etc. in the Tree model.

app/views/attendances/_row.html.erb
```diff
@@ -1,8 +1,8 @@
 <td><%= avatar(attendance.user) %></td>
 <td><%= link_to attendance.user.name, group_membership_path(attendance.event.group, attendance.user) %></td>
-<% ["Absent", "Tardy", "Present"].each_with_index do |status, i| %>
+<% [nil, 0, 1, 2].each_with_index do |status| %>
```
mattscilipoti: Next week, I will have no idea what these number represent.  We need an enumeration where the text is associated with the numbers.  `Attendance::States = ["Absent", "Tardy", "Present"]` (or an instance method)?

app/views/attendances/_show_table.html.erb
```diff
@@ -0,0 +1,12 @@
+<table>
+  <tr>
+    <th>Title</th>
+    <th>Group</th>
+    <th>Event Date</th>
+    <th>Date Taken</th>
+    <th>Status</th>
+  </tr>
+  <% attendances.each do |attendance| %>
+  <%= render partial: "attendances/show_row", locals: {attendance: attendance, event: attendance.event} %>
```
mattscilipoti: recommend indentation.

app/models/membership.rb
```diff
@@ -86,4 +78,15 @@ def has_descendants?
...
+    end
+  end
```
mattscilipoti: Are we removing ancestor memberships on removal?

RobertAKARobin: No

app/views/users/show.html.erb
```diff
@@ -77,5 +47,12 @@
 </section>
 
 <section>
-  <h3>Project Feedback</h3>
+  <h3>Profile</h3>
+  <%= form_tag user_path(@user), method: :patch do %>
+  <%= render partial: "form" %>
```
mattscilipoti: No need to extract the "form_tag".  Use `form_for` and rails will automatically manage the patch/post using `@user.new_record?`

app/views/attendances/_row.html.erb
RobertAKARobin: Will add

app/views/attendances/_counts.html.erb
mattscilipoti: Ahh.  It's always painful when get in that position, but sometimes it's necessary.

PR #142, opened 2015-10-22
removing excess stuff (CAUTION: See below)

RobertAKARobin:


RobertAKARobin:
CAUTION: This PR contains a migrate that will delete a lot of data. May be a good idea to back-up before running this.

mattscilipoti:
It does not look like you are "removing excess stuff” in this PR. I see many additions.  It looks like is_priority is a pretty important concept in that PR.  I can’t see how create/destroy parent memberships is related to isPriority.

RobertAKARobin:
It isn't related. But that has to be included because of a prior PR which included a migration that created the field. The PR was rolled back; the migration was not.

RobertAKARobin:
Beyond is_priority, the only addition I see is the new migration file.

mattscilipoti:
I stand corrected.  The other "addition" was the migration... which, btw, doesn't remove any files.

is_priority:
All the more reason why that migration should be in a separate PR.  To clearly identify, and comment on, that special situation.

RobertAKARobin:
It doesn't remove files. It removes data in the DB.

RobertAKARobin:
I meant that this PR will fail tests due to "pending migrations" until the is_priority migration is run.

RobertAKARobin:
And Matt, I totally agree with you on keeping things discreet. I let myself get out of control by patching things willy-nilly. The discussion Jesse and I had yesterday was whether it would be a better investment of our time to go through and separate out the big pull request into mini pull requests, or just treat is as a lesson learned and move forward.

PR #143, opened 2015-10-22
added is-priority

RobertAKARobin:


RobertAKARobin:
This will be used as a way of identifying which groups are "squads".

PR #144, opened 2015-10-22
remove excess memberships, and attendances/submissions/observations from instructors

RobertAKARobin:


jshawl:
@RobertAKARobin this branch has conflicts. Also, do you mind pairing with @mattscilipoti about this pr? Nervous about the deletions

db/migrate/20151022183954_remove_excess.rb
```diff
@@ -0,0 +1,15 @@
...
+  def change
+    Group.at_path("ga").memberships.destroy_all
+    Group.at_path("ga_wdi").memberships.destroy_all
+    Group.at_path("ga_wdi_dc").memberships.destroy_all
+
+    instructors = ["adambray", "ebirving", "andrewsunglaekim", "amaseda", "jmas13", "nolds9", "mattscilipoti", "jshawl", "robertakarobin"]
+    instructors.each do |username|
+      user = User.named(username)
+      user.attendances.destroy_all
+      user.observations.destroy_all
```
mattscilipoti: Does this remove observations about us, or observations performed by us?

db/schema.rb
```diff
@@ -58,8 +58,9 @@
     t.integer  "group_id"
     t.integer  "user_id"
     t.boolean  "is_admin"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
+    t.datetime "created_at",  null: false
+    t.datetime "updated_at",  null: false
+    t.boolean  "is_priority"
```
mattscilipoti: Where is the "db/schema.rb" change for this migration?

db/migrate/20151022183954_remove_excess.rb
RobertAKARobin: About us. We shouldn't have any

db/schema.rb
RobertAKARobin: There isn't one, because the migration doesn't change the schema at all -- it just deletes records.

RobertAKARobin:
Conflicts were due to previous merge. Fixed.

PR #148, opened 2015-10-22
moved spinner icon into own public file

RobertAKARobin:


RobertAKARobin:
Would it be more memory-efficient (albeit less cool) to use a GIF?

PR #149, opened 2015-10-22
submissions save status on load from Github so that averages/totals may be calculated

RobertAKARobin:


RobertAKARobin:
@jshawl Do submissions check who closed the PR/issue? Would it still be considered complete if a student closed the PR/issue?

app/helpers/application_helper.rb
```diff
@@ -97,4 +97,17 @@ def markdown(text)
     return text
   end
 
+  def gh_submission submission
+    gh_info = submission.github_pr_submitted
+    if gh_info
+      if gh_info[:state] == "closed"
```
jshawl: yeah technically a student can close and it will count as reviewed but i think that's ok, because we don't count reviews other than to give students the feedback they want

PR #153, opened 2015-10-23
path is sorta cached

RobertAKARobin:


app/models/group.rb
```diff
@@ -19,6 +19,8 @@ class Group < Tree
   before_save :validate_uniqueness_of_title_among_siblings
   after_save :update_child_members
 
+  attr_accessor :path
```
mattscilipoti: This is unnecessary, since you override it with a defined method on line:56.

mattscilipoti:
:+1:  Standard ruby memoization pattern.

app/models/group.rb
RobertAKARobin: Resolved

PR #154, opened 2015-10-23
added squad toggler to user profile

RobertAKARobin:


app/controllers/users_controller.rb
```diff
@@ -21,7 +21,7 @@ def show
     end
     @is_current_user = (@user.id == current_user.id)
     @is_editable = @is_current_user && !@user.github_id
-    @memberships = @user.memberships
+    @memberships = @user.memberships.sort{|a,b| a.group.path <=> b.group.path}
```
mattscilipoti: Acceptable, but I recommend using @user.memberships.sort_by(:group_path) and delegating group_path (or just path) to self.group (law of demeter).

app/models/group.rb
```diff
@@ -70,6 +70,14 @@ def subnonadmins
     self.descendants_attr("memberships").select{|m| !m.is_admin}.collect{|m| m.user}.uniq.sort{|a,b| a.last_name <=> b.last_name}
   end
 
+  def member user
+    memberships = self.memberships
+    if user.class <= String
```
mattscilipoti: <= String? That's different.  Is this a special ruby comparison?

app/models/group.rb
```diff
@@ -70,6 +70,14 @@ def subnonadmins
...
+      user = User.named(user)
+    end
+    return memberships.find_by(user: user)
+  end
```
mattscilipoti: method "member" returns collection of memberships?  Name seems inappropriate.  How about named scope on memberships: `for _user`.  Then this would be `group.memberships.for_user(user)`.

app/views/users/show.html.erb
```diff
@@ -6,9 +6,23 @@
...
+    <p>None yet!</p>
   <% else %>
-  <ul class="tree"><%= membership_list(@memberships) %></ul>
+    <ul>
+    <% @memberships.each do |membership| %>
+      <li class="links">
+        <span class="breadcrumbs">
+          <%= breadcrumbs(membership.group) %>
+        </span>
+        <%= "<a>admin</a>".html_safe if membership.group.admins.include?(@user) %>
```
mattscilipoti: An anchor tag without an href?  Maybe your looking for `link_to_if`?  If you always want the text, but only want the link in some conditions.  Same for empty "squad" a tag below.

app/models/group.rb
RobertAKARobin: It compares the classes. "If this class inherits from or is equal to"

app/models/group.rb
RobertAKARobin: It will only return a single membership, hence `find_by`

app/views/users/show.html.erb
RobertAKARobin: It's to make the CSS work without needing to create additional selectors.

app/controllers/users_controller.rb
RobertAKARobin: `path` is a method, not a field. Will that still work?

app/models/group.rb
RobertAKARobin: This has been moved to https://github.com/ga-dc/garnet/pull/159

PR #155, opened 2015-10-23
added path back to group model

RobertAKARobin:


RobertAKARobin:
The intent is to keep it cached

PR #156, opened 2015-10-23
membership should not create parent and child memberships

RobertAKARobin:


mattscilipoti:
Did I miss the commit that made these obsolete?

RobertAKARobin:
They've been obsolete all along, actually -- we haven't done anything with parent and child memberships.

PR #164, opened 2015-10-27
(WIP) Report Card

jshawl:


app/controllers/memberships_controller.rb
```diff
@@ -28,9 +28,10 @@ def destroy
 
   def show
     @group = Group.at_path(params[:group_path])
-    @is_admin = @group.admins.include?(current_user)
+    cur_user = User.find(session[:user]["id"])
```
jshawl: cc @RobertAKARobin `current_user` was returning nil here so here's a bandaid solution

PR #165, opened 2015-10-27
debug: revert to Webrick for dev.

mattscilipoti:
- sort gems
- had difficulties getting unicorn to support pry/debugger, so reverted to webrick for dev.
- comment out “spring”


jshawl:
:+1:

PR #170, opened 2015-10-28
further enhancements to front-endiness of loading GH issues and updating submissions

RobertAKARobin:


RobertAKARobin:
Further enhances the front-endiness of loading issues.

app/assets/javascripts/application.js
```diff
@@ -13,3 +13,15 @@
 //= require jquery
 //= require jquery_ujs
 //= require_tree .
+
+
+function update(url, object){
```
jshawl: would be useful to add a callback here as function argument, which is invoked in promise, or return promise

app/assets/javascripts/assignments.js
```diff
@@ -35,3 +35,42 @@ $(function(){
     })
   }
 })
+
```
jshawl: does the below code make the above code obsolete?

app/models/assignment.rb
```diff
@@ -50,4 +50,12 @@ def api_repo_issues_url
     url += "/issues?state=all"
   end
```
jshawl: does the below code make the above code obsolete?

app/views/assignments/show.html.erb
```diff
@@ -5,13 +5,12 @@
 
 <section>
 <h3>Submissions</h3>
-<p>
-  <%= @submissions_percent_complete %>% complete.
-</p>
+<p><%= percent_of(@submissions, 2) %>% complete</p>
+<button onclick="loadIssues('<%= assignment_issues_path(@assignment) %>')">Count issues</button>
```
jshawl: does this need to be an AJAX call? why not just count from the controller?

jshawl:
I'm a little bummed about having to mark something complete here, and then as well on github. There is no visual representation of whether an issue is present or not.

app/assets/javascripts/application.js
RobertAKARobin: Good call

app/models/assignment.rb
RobertAKARobin: No; `#issues_url` just returns the URL at which the issues can be found

RobertAKARobin:
There is! This introduces a "Count issues" button, which, upon clicking, replaces the text of all the "View issues" links with the number of "open" and "closed" issues for each user (see image).

screen shot 2015-10-28 at 9 57 18 am

The advantage of doing it this way is that the user may not need to see the counts of all the issues, so the AJAX call only fires if they need it. Marking a submission as "complete" and closing issues are still separate processes, but I think that's good because otherwise a student could mark all their homework as "complete" simply by opening an issue and closing it, and they may have multiple issues/PRs on one repo.

jshawl:
I'm getting an off by one error

RobertAKARobin:
OK, will investigate

app/assets/javascripts/assignments.js
RobertAKARobin: I guess so, but I felt nervous and so didn't want to remove it.

RobertAKARobin:
@jshawl Responded to your comments; thanks! I think this should fix it.

jshawl:
last two are showing issues, but only the last one has an issue

PR #180, opened 2015-10-29
removed errors controller

RobertAKARobin:


RobertAKARobin:
...because it doesn't do anything anywhere.

PR #184, opened 2015-10-29
add new relic: gem, config/newrelic.yml

mattscilipoti:
- Uses my personal key.  If we like it, we should  get GA account.

### DIFF
--- a/config/newrelic.yml
+++ b/config/newrelic.yml
@@ -1,0 +1,52 @@
+#
+# This file configures the New Relic Agent.  New Relic monitors Ruby, Java,
+# .NET, PHP, Python and Node applications with deep visibility and low
+# overhead.  For more information, visit www.newrelic.com.
+#
+# Generated October 29, 2015
+#
+# This configuration file is custom generated for General Assembly_24
+#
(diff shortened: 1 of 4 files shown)


config/newrelic.yml
```diff
@@ -0,0 +1,52 @@
...
+#
+# For full documentation of agent configuration options, please refer to
+# https://docs.newrelic.com/docs/agents/ruby-agent/installation-configuration/ruby-agent-configuration
+
+<% app_name = "GArnet" %>
+
+common: &default_settings
+  # Required license key associated with your New Relic account.
+  # TODO: Matt's account.  Convert to "GA" account
+  license_key: d8ead2cc15b31e3649f3d8071ba9273b87a971a7
```
jshawl: should this be gitignored?

mattscilipoti: Good question.  I should have added to figaro.  Doesn't matter just yet though.  This is just for demo.  It's a free account.  We'll be closing that account when we make a decision.  Then we'll add the key figaro and use ENV setting - file will not be gitignored.

PR #185, opened 2015-10-29
slightly refactored tree methods and added some user methods

RobertAKARobin:


app/models/membership.rb
```diff
@@ -23,7 +23,7 @@ def make_admin_nonadmin
   end
 
   def self.extract_users array
-    array.collect{|m| m.user}.uniq.sort{|a,b| a.name <=> b.name}
+    array.collect{|m| m.user}.uniq.sort_by(&:name)
```
jshawl: :+1:

RobertAKARobin: :+1: for Twoproc Shakur @adambray

PR #186, opened 2015-10-29
removed default values for submissions and attendances so can be nil

RobertAKARobin:


RobertAKARobin:
This allows us to show that an attendance/submission has *not* been taken. Currently, they default to "absent" by default, which makes it difficult to see which absences are "real".

jshawl:
:+1: waiting for today's attendance to be taken before merge

RobertAKARobin:
It shouldn't affect it, but good call.

jshawl:
i feel similarly! just in case...

PR #187, opened 2015-10-29
Date to string methods

RobertAKARobin:


app/models/event.rb
```diff
@@ -3,6 +3,7 @@ class Event < ActiveRecord::Base
   has_many :attendances, dependent: :destroy
   has_many :users, through: :attendances
   validates :date, presence: true
+  validates :title, presence: true
```
jshawl: should this have a corresponding view update to show errors if a validation fails?

RobertAKARobin:
It does automatically because of the global error catcher.

app/models/event.rb
RobertAKARobin: It does automatically because of the global error handler.

PR #189, opened 2015-10-29
Better helpers

RobertAKARobin:


RobertAKARobin:
Was going to make the group methods a separate PR, but they're used in the subsequent commit.

PR #194, opened 2015-10-29
consolidated lots of pages and partials and made everything rails-y-er

RobertAKARobin:


RobertAKARobin:
What the eff.

app/controllers/errors_controller.rb
```diff
@@ -0,0 +1,8 @@
+class ErrorsController < ApplicationController
```
jshawl: wasn't this removed in https://github.com/ga-dc/garnet/commit/3f51361a635036aaa5c54cb726fff13ea6051e50

PR #195, opened 2015-10-29
consolidated many views and made things (esp forms) rails-y-er

RobertAKARobin:


RobertAKARobin:
@jshawl I THINK this fixes one-click grading. It looks like a lot because I copied all the partial views that were used only once into the "full" views.

PR #198, opened 2015-10-29
consolidated events controller and views

RobertAKARobin:


app/controllers/events_controller.rb
```diff
@@ -11,13 +11,12 @@ def index
...
     @group = Group.at_path(params[:group_path])
     @event = @group.events.new(event_params)
-    date = params[:date]
     date = DateTime.new(
-      date[:year].to_i,
-      date[:month].to_i,
-      date[:day].to_i,
-      date[:hour].to_i,
-      date[:minute].to_i
+      event_params["date(1i)"].to_i,
```
jshawl: can we use a date field or something? this feels code-smelly

jshawl:
:+1: pending comment

RobertAKARobin:
Sure! But I'm not sure how... Is there an alternative to `<select>` date menus?

jshawl:
what about

```html
<input type="hidden" value="unix time stamp">
```

then in params parse to datetime, from beginning of day

RobertAKARobin:
I dunno... That'll be just as much business logic as with this `datetime`, only in Javascript. I'd rather trust Rails / user what's ready.

PR #201, opened 2015-10-29
consolidated users views and controllers

RobertAKARobin:


app/helpers/application_helper.rb
```diff
@@ -165,4 +162,22 @@ def status_buttons record
     return output.html_safe
   end
```
jshawl: when would you use `records_accessible_by` and `td_averages`?

app/views/users/show.html.erb
```diff
@@ -3,20 +3,94 @@
   <h2 title="user#<%= @user.id %>"><%= @user.username %></h2>
 </div>
 
-<% if is_admin_of_anything?(@user) %>
```
jshawl: admin only, right?

config/routes.rb
```diff
@@ -7,9 +7,11 @@
   post '/sign_in', to: 'users#sign_in!'
   get '/sign_out', to: 'users#sign_out', as: :sign_out
 
-  get "github/authorize", to: "users#gh_authorize", as: :gh_authorize
-  get "github/authenticate", to: "users#gh_authenticate", as: :gh_authenticate
-  get "github/refresh", to: "users#gh_refresh", as: :gh_refresh
+  scope :github do
```
jshawl: :star:

RobertAKARobin:
`records_accessible_by` is actually also defined in the User model, which is used in the Users controller. I forgot I started with it as a helper. I can delete that.

`td_averages` is just used on the Groups show page. It's used only once, but there's so much logic involved that I didn't want to put it in the view or controller.

app/views/users/show.html.erb
RobertAKARobin: As opposed to...?

jshawl: Sorry I meant shouldn't the groups be visible to admins only? I.e put back this line?

RobertAKARobin: I put it back in because I thought it would be nice for students to see in Garnet what their squads and classrooms are. I'd like to keep it.

RobertAKARobin:
Is it OK if I delete `records_accessible_by` in a subsequent commit?

PR #221, opened 2015-11-03
Dont show members section by default

jshawl:


RobertAKARobin:
This just has them collapsed by default?

The folding is all done via jump tags and CSS, so JS shouldn't be necessary. The JS is just in there to unfold memberships by default.

RobertAKARobin:
@jshawl You'd rather have everything collapsed by default?

The folding is all done via jump tags and CSS, so JS isn't necessary. This JS is just in there to unfold the Memberships section by default if there's no hash in the URL.

jshawl:
@RobertAKARobin this pull request removes the default opening of members, and then allows the user to collapse a section (which requires JS to update the hash)

RobertAKARobin:
It's done with the `.fold` class and jump anchors:

https://github.com/ga-dc/garnet/blob/master/app/views/groups/show.html.erb#L2

I kinda prefer that to JS, personally.

RobertAKARobin:
Oh, I get it, so that they can collapse a section by clicking on the jump anchor when it's unfolded, rather than having to manually remove the hash from the URL or something?

jshawl:
yes

RobertAKARobin:
:+1:

PR #227, opened 2015-11-05
reversed order of events on group page and gave todays events a background color, fixed #228

### LINKED ISSUE #228: Sort attendance reverse order - latest at the top.
We rarely are interested in an earlier attendance.

RobertAKARobin:


RobertAKARobin:
@mattscilipoti Can you :+1: ?

app/views/groups/show.html.erb
```diff
@@ -76,7 +76,7 @@
       </tr>
       <% @events.each do |event| %>
         <% attendances = @attendances.select{|a| a.event == event} %>
-        <tr>
+        <tr <%= "style='background-color:#9cf;'".html_safe if event.date.today? %>>
```
mattscilipoti: Is today's date important in other situations too? If so, how about a convention for today's date (data attribute and css)?  Then, it will be highlighted in other situations too.
Also, this feels like a total hack (inline style, hard coded unique color, html_safe).

mattscilipoti: I'll accept it if we schedule time, today, to correct.

RobertAKARobin: Fair enough. Fixed; refresh to see change.

mattscilipoti:
:+1: after above changes

PR #229, opened 2015-11-05
add link to filter attendances

jshawl:
To make taking attendance easier, there is a now a link that allows the user to filter attendances where the status is nil.

At first, I thought about doing this in JS, but the url params will allow us to send the filter link to others.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -19,7 +19,7 @@ class EventsController < ApplicationController
     @event = Event.find(params[:id])
     @current_user_is_admin = @event.group.has_admin?(current_user)
     redirect_to(current_user, flash:{alert: "You're not authorized."}) if !@current_user_is_admin
-    if params[:filter] == "na"
+    if params[:status] == "nil"
       @attendances = @event.attendances.where(status:nil)
     else
       @attendances = @event.attendances
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -19,7 +19,11 @@ class EventsController < ApplicationController
     @event = Event.find(params[:id])
     @current_user_is_admin = @event.group.has_admin?(current_user)
     redirect_to(current_user, flash:{alert: "You're not authorized."}) if !@current_user_is_admin
-    @attendances = @event.attendances
+    if params[:filter] == "na"
+      @attendances = @event.attendances.where(status:nil)
+    else
+      @attendances = @event.attendances
+    end
     if params[:group]
       @group = Group.at_path(params[:group])
       @attendances = @attendances.select{|a| a.user.is_member_of(@group)}
--- a/app/views/events/show.html.erb
+++ b/app/views/events/show.html.erb
@@ -2,10 +2,10 @@
 
 
 <p>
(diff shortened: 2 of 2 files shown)


app/controllers/events_controller.rb
```diff
@@ -19,7 +19,11 @@ def show
     @event = Event.find(params[:id])
     @current_user_is_admin = @event.group.has_admin?(current_user)
     redirect_to(current_user, flash:{alert: "You're not authorized."}) if !@current_user_is_admin
-    @attendances = @event.attendances
+    if params[:filter] == "na"
```
jshawl: feels a little smelly. myabe `params[:status] == "nil"`?

app/views/events/show.html.erb
```diff
@@ -1,4 +1,14 @@
 <h2><%= "#{@event.title}, #{@event.date_s}" %></h2>
+
+
+<p>
+<% if request.url.match(/filter=na/) %>
+  <%=link_to "Show All", request.url.gsub(/filter=na/,'') %>
+<% else %>
+  <%=link_to "Filter by N/A", request.url + "&filter=na" %>
```
jshawl: will break if there isn't a `?` before

app/controllers/events_controller.rb
RobertAKARobin: Yeah, I'd prefer `params[:status]`. May as well match the name of the field.

RobertAKARobin: Also, I'd always forget whether `?filter=na` filters *on* `na` or filters *out* `na`.

app/views/events/show.html.erb
```diff
@@ -1,4 +1,14 @@
 <h2><%= "#{@event.title}, #{@event.date_s}" %></h2>
+
+
+<p>
+<% if request.url.match(/filter=na/) %>
```
RobertAKARobin: Use `params[:status]` here? RegEx seems a lil clunky

app/views/events/show.html.erb
RobertAKARobin: How about `link_to "Filter by N/A", event_path(event, status: "na")`? http://stackoverflow.com/a/2696001/2053389

jshawl:
@RobertAKARobin whachu think now? https://github.com/ga-dc/garnet/commit/d1cdefd105bf38cb5ee9c9303bbacaf7cb799a27

app/views/events/show.html.erb
```diff
@@ -1,4 +1,14 @@
 <h2><%= "#{@event.title}, #{@event.date_s}" %></h2>
+
+
+<p>
+<% if params[:status] == "nil" %>
+  <%=link_to "Show All", request.url.gsub(/&status=nil/,'') %>
+<% else %>
+  <%=link_to "Filter by N/A", request.url + "&status=nil" %>
```
RobertAKARobin: How about `link_to "Filter by N/A", event_path(event, status: "na")`? `link_to` has a built-in querystring-adder.  http://stackoverflow.com/a/2696001/2053389

jshawl: effff yteaaaah just wat i was lookin fo

PR #236, opened 2015-11-06
hotfix: ensure migration works locally

mattscilipoti:
- RemoveExcessMemberships was adding users, that didn't exist locally, to groups.

Note: this should not affect Production (please think about that too).  In fact, this migration has already run successfully on production.


jshawl:
:+1:

RobertAKARobin:
:+1:

PR #238, opened 2015-11-10
hide form if attendance event already exists for today; fixes #237

### LINKED ISSUE #237: When attempting to create a new event, warn the user if an event already exists for the day.
An alert box or something.

jshawl:


RobertAKARobin:
:+1: 

We'll have to trim down those controllers at some point, though.

PR #241, opened 2015-11-11
redirect to existing event if one was created within 500 seconds

jshawl:


jshawl:
cc @mattscilipoti

RobertAKARobin:
Might use `created_at` instead?

app/controllers/events_controller.rb
```diff
@@ -11,6 +12,10 @@ def create
       event_params["date(5i)"].to_i
     )
     @event.date = date
+    if @event.date.to_i - last_event.date.to_i < 500
+      redirect_to last_event
```
RobertAKARobin: Can also `return redirect_to last_event`, but whatevs.

app/controllers/groups_controller.rb
```diff
@@ -14,7 +14,7 @@ def show
     @assignments = @submissions.collect(&:assignment).uniq.sort_by(&:due_date)
     @attendances = @group.members_attendances.uniq
     @events = @attendances.collect(&:event).uniq.sort_by(&:date).reverse
-    @event_for_today_already_exists = @events.first.date.strftime("%Y%m%d") == DateTime.now.strftime("%Y%m%d")
+    @event_for_today_already_exists = @events[0] ? @events.first.date.strftime("%Y%m%d") == DateTime.now.strftime("%Y%m%d") : false
```
RobertAKARobin: Was throwing an error if the group had no events.

mattscilipoti: `@event_for_today_already_exists = @events.any? { |event| event.today? }`

mattscilipoti: Prefer this queries db directly.  Looking now.

mattscilipoti:
```
class Event < AR::Base
  def self.today
    where("date >= ?", Time.zone.now.beginning_of_day)
  end

  def self.today?
     today.any?
  end
end
```
`@event_for_today_already_exists = @events.today?`

Issue #243, opened 2015-11-11
Percent Complete and N/As


dan-ator:
I think that the assignments listed as N/A (meaning assigned, turned in, but not yet graded) should not be included in the calculation for percent complete.  It's a small point, but I don't like seeing the percent complete so low on my "report card" when in fact I did complete them (or they aren't yet due).

jshawl:
@dan-ator great point!!! Are you interested in contributing? Maybe `reject` those submissions that have a status of 0 ("n/a") - https://github.com/ga-dc/garnet/blob/174a1e4b611b60fc720e7e430f1fb36800fa2631/app/controllers/users_controller.rb#L29

RobertAKARobin:
Fixed by https://github.com/ga-dc/garnet/pull/256

Issue #240, opened 2015-11-11
Allow manual setting of assignment due dates


ebirving:
At the moment, it defaults to the date the assignment is created, but that's contrary to the "HW should go in Garnet as soon as it's assigned" mandate.

RobertAKARobin:
Fixed by https://github.com/ga-dc/garnet/pull/256

PR #256, opened 2015-11-18
manual assignment of due dates for assignments

ebirving:
Implements a number of changes to address https://github.com/ga-dc/garnet/issues/240 and https://github.com/ga-dc/garnet/issues/243
1. Adds date select field to form for assignments. Default remains current date, but this allows for creation of assignments in advance.
2. Removes submissions with status N/A from calculation of percent complete in student report card view.
3. Hides submissions from admin's to-do list until the due date. All submissions, regardless of due date or status, will still be visible on the assignment's show page, allowing access for early grading if applicable.


RobertAKARobin:
:+1: Niiiiice.

Issue #176, opened 2015-10-28
Photo on profile page extends above top border of window when expanded on hover


ebirving:

RobertAKARobin:
Fixed by https://github.com/ga-dc/garnet/pull/257

PR #257, opened 2015-11-18
remove hover effect from welcome screen avatar

ebirving:
Addresses https://github.com/ga-dc/garnet/issues/176


RobertAKARobin:
:+1:

PR #258, opened 2015-11-18
added ancestry gem and updated files accordingly

RobertAKARobin:
Only real file changes were changing `path` to `path_s` in several places.


andrewsunglaekim:
how does this PR effect the groups we currently have in our DB, if at all?

RobertAKARobin:
It shouldn't affect them at all. I'm continuing to test.

app/models/group.rb
```diff
@@ -38,19 +36,19 @@ def validate_uniqueness_of_title_among_siblings
...
       next if i == 0
-      group = group.children.find_by(title: path[i])
+      group = group.children.find_by(title: title)
     end
     return group
   end
 
-  def path
-    @path ||= self.ancestors.collect{|g| g.title}.join("_")
+  def path_s
```
adambray: Is there a more descriptive name for this method? It took me a while to get this was a path_string...

app/models/group.rb
```diff
@@ -106,6 +104,15 @@ def is_childless?
     return self.children.count < 1
   end
 
+  def create_descendants hash, name
```
adambray: Where is this method used?

app/models/group.rb
RobertAKARobin: We could call it `path_string`.

app/models/group.rb
RobertAKARobin: Only in seeds.

andrewsunglaekim:
as long as thats the case, im good with it :+1:  Maybe a good medium ground that satiates the test-coverage is to make tests for groups that pass with either setup?

app/models/group.rb
adambray: I'd vote for that, personally

app/models/group.rb
adambray: got it... do you have an updated seed script for this? I noticed it's not in the PR

adambray:
Currently checking this branch out and testing the migrations on production data locally

app/models/group.rb
RobertAKARobin: That file's unaffected by this. `#create_descendants` was in the Tree model, which is now unused (and will be deleted in a subsequent push), so I moved it to the Group model.

adambray:
So this isn't necessarily a functional change, but after switching to this branch, migrating, and running the app on production data, I noticed that group paths look different:
screen shot 2015-11-18 at 2 29 58 pm

RobertAKARobin:
@adambray Yeah, I'm trying to figure that out right now

adambray:
@RobertAKARobin also getting an error on this branch that I'm not getting in prod, when you visit a student's 'show' page:

screen shot 2015-11-18 at 2 32 21 pm

PR #260, opened 2015-11-18
added ancestry gem and updated files accordingly, fixes #230

### LINKED ISSUE #230: Implement the Ancestry gem and stop trying to reinvent the wheel with the Tree model, Robin, you stubborn fool.

RobertAKARobin:
Only real file changes were changing `path` to `path_string` in several places.

Also, reproduced existing `group#ancestors` method in the migration file itself since this method will be replaced by Ancestry's, but is necessary to convert to Ancestry in the first place.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ gem 'pg'
 
 # >>>>>>>>>>>>>>>>>>>>>>>>>
 # remaining gems are sorted alphabetically
+gem 'ancestry'
 gem 'bcrypt' # Use ActiveModel has_secure_password
 gem 'cancancan'
 # Use CoffeeScript for .coffee assets and views
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -37,6 +37,8 @@ GEM
       thread_safe (~> 0.3, >= 0.3.4)
       tzinfo (~> 1.1)
     addressable (2.3.8)
+    ancestry (2.1.0)
+      activerecord (>= 3.0.0)
     arel (6.0.3)
     bcrypt (3.1.10)
     binding_of_caller (0.7.2)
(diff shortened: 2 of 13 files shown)


adambray:
This seems to address the two previous issues I found (incomplete path names and errors on user show page).

That said, just want to make sure the following is expected... looking at memberships, they seem a bit inconsistent. This may be just caused by our current Prod data, and not this code, but I noticed that all students are in:
GA
GA_WDI_DC_7
GA_WDI_DC_7_<SQUAD>

screen shot 2015-11-18 at 3 38 34 pm

----

But instructors are not in GA, just 7 and below:
screen shot 2015-11-18 at 3 40 51 pm

adambray:
@RobertAKARobin items marked todo appear to be broken, both grades and attendance on my home page don't show me members of my squad who have items that are `n/a`

adambray:
@RobertAKARobin another change: when making an observation, in the new code, I see one option for group: `ga_wdi_dc_7_adam`, in production, I see two, the former one, plus just `ga_wdi_dc_7`.

adambray:
I can't test this to compare with prod, but another issue:

1. Go to the group page for ...wdi_7
2. Click members
3. Click remove next to a member's name
4. Refresh, note that the member is still listed
5. Try to remove again, get an error `undefined method `destroy!' for nil:NilClass`
6. Refresh, note member is still listed
7. Click user's name, note that all information is still present

adambray:
Note you get different behavior when removing from a sub-group like the squad group. 

* If you remove a user from from a parent group AND all subgroups, they finally disappear in the parent group.
* If you remove a user from a subgroup but not the parent group, they disappear in the subgroup, but are still listed in the parent group (expected)
  * however, if you then click on the user's name in the parent group, you get an error

screen shot 2015-11-18 at 4 01 04 pm

RobertAKARobin:
@adambray Fixed the removing error. It was because the page still shows folks even if they're not a member of *that* group but rather a *sub*group. Now it will still show them, but without a remove button.

PR #264, opened 2015-11-19
Faster sql

RobertAKARobin:


adambray:
Sorting by last name is broken on "ToDo" sections, both attendance and HW Grades.

adambray:
Order of dropdown for group when making observations is reversed

adambray:
Sorting by last name is broken on assignments show.

adambray:
Sorting appears to be last, then first on attendance show page. I think this is an improvement, but I'd be curious to know why that changed!

RobertAKARobin:
@adambray Just pushed a commit to fix sorting.

RobertAKARobin:
I don't really think the group order for observations is an issue... I'd kinda prefer that it be on `ga_wdi_dc_7`. Which order we like best is likely to be something we'd end up having a big conversation about, and since it doesn't really impact performance I'd vote we just leave as-is. I didn't make any changes to the `users#show` page at any rate.

adambray:
@RobertAKARobin agreed that default group in observation doesn't matter, but I just wanted to point it out.

Sorting in members list on group#show is fixed.

Sorting by last name in ToDos is still broken

adambray:
never mind! that's not a regression. it's an issue in the current prod version

adambray:
I think this is good to merge! :+1:

andrewsunglaekim:
:+1:  noice i like the new helpers

PR #284, opened 2015-12-02
Refactor routes and controller actions surrounding users and sessions…

adambray:
…. also removes un-used actions and methods on Github


adambray:
Would appreciate some manual testing on all user related actions on this one. I did testing myself, but there's a lot moved around here, so it's worth some extra efforts.

Also, apologies, I tried to limit the scope on this PR, but a lot of stuff was all tied together, and I really wanted to clean up some of this code.

app/controllers/observations_controller.rb
```diff
@@ -1,7 +1,8 @@
 class ObservationsController < ApplicationController
 
   def create
-    @user = User.named(params[:user_user])
+    # TODO: figure out how to fix the weirdness in this route -AB
+    @user = User.named(params[:user_user] )
```
mattscilipoti: Why?

spec/requests/installfest_spec.rb
```diff
@@ -2,18 +2,18 @@
 
 RSpec.describe "Installfest", :type => :request do
   it "requires un-authenticated access to /users/is_registered endpoint" do
-    get "/users/is-registered"
+    get "/users/anything/is_registered"
```
mattscilipoti: Note! These changes force changes in InstallFest.  This spec was here to indicate what InstallFest was expecting.

mattscilipoti: I'm ok with either route.  I didn't mind this being a little obscure.  As I was writing it, I wasn't sure I liked and endpoint to check to see what github users were registered on our app.  Didn't see another answer to support InstallFest.

mattscilipoti:
I gave a cursory review.  It's too late to inspect all the guts in the middle.

RobertAKARobin:
Major props, Adam! The huge Users controller has been making me cringe for weeks.

Concern: I feel like we should wait until there are sufficient tests to merge this in. Counterpoint: We don't really have tests now so it doesn't make much difference, and also I worry about testing slowing down development. If we decide we do need tests, I'd love to pair with someone on writing them!

app/controllers/observations_controller.rb
adambray: That's right. I'll push a small commit to fix the space.

spec/requests/installfest_spec.rb
adambray: There is an associated PR in installfest to account for these changes!

As for the routes, I feel like obscurity isn't really security, and to me, one isn't more obscure than the other. 

If we really want to make this secure, we should probably use a shared secret between the client and the server. I.e. check for the presence of some token here in the URL (matched to an ENV var), and then give students that token string during installfest.

adambray:
Yeah, I want tests too. Gonna try to do that today.

PR #286, opened 2015-12-03
removes show/hide pics for simplicity, separates users into 3 tables …

adambray:
…on group view, refactors groups#nonadmins to be more SQL-y

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -1,21 +1,30 @@
 class GroupsController < ApplicationController
 
   def show
-    @hide_pics = (params[:show_pics] ? false : true)
     if params[:path]
       @group = Group.at_path(params[:path])
     else
       @group = Group.first
     end
     @is_admin = @group.has_admin?(current_user)
-    @nonadmins = @group.nonadmins.to_a
+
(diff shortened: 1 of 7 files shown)


app/controllers/groups_controller.rb
```diff
@@ -1,21 +1,30 @@
...
     else
       @group = Group.first
     end
     @is_admin = @group.has_admin?(current_user)
-    @nonadmins = @group.nonadmins.to_a
+
+    @nonadmins = @group.nonadmins
+    @active_members    = @nonadmins.select{ |u| u.memberships.find_by(group: @group, status: Membership.statuses[:active]) }
+    @inactive_members  = @nonadmins.select{ |u| u.memberships.find_by(group: @group, status: Membership.statuses[:inactive]) }
+    @subgroup_members  = @nonadmins.select{ |u| !u.memberships.find_by(group: @group) }
```
RobertAKARobin: Maybe save all the memberships to a variable and then iterate through them, instead of getting all the memberships 3 times? #pedantic

RobertAKARobin:
I'd say merge it in as long as it doesn't noticeably slow down execution time. We can always refactor later.

Issue #279, opened 2015-11-30
Incorrect Date on Attendance Event Submission


jocelynhoule:
When adding an event for attendance, resulting submission shows today's date only for students.  Steps are:
>Add Item Title
>Set date to 11/20/15
>Save Event
>expected result would be class roster for 11/20
>actual result is class roster for today's date 11/30
[garnet_issue.pptx](https://github.com/ga-dc/garnet/files/47867/garnet_issue.pptx)

mattscilipoti:
Log stuff.  Looks like Controller is getting the correct params.
```
I, [2015-11-30T16:52:03.046018 #13092]  INFO -- : Started POST "/groups/ga_pmi_dc_1/events" for 50.200.196.50 at 2015-11-30 16
:52:03 -0500
I, [2015-11-30T16:52:03.053334 #13092]  INFO -- : Processing by EventsController#create as HTML
I, [2015-11-30T16:52:03.053756 #13092]  INFO -- :   Parameters: {"utf8"=>"✓", "authenticity_token"=>"IV1v/nf/zG5tVzi8e2k2NImKN
QN22DeqxmkNXjyhX+8Ytu12Ch/ehSuhzSLiXbtAlBjAw7HjF/J64bP+SQRSLA==", "event"=>{"title"=>"Week 3 Class - 11", "date(1i)"=>"2015",
"date(2i)"=>"11", "date(3i)"=>"20", "date(4i)"=>"16", "date(5i)"=>"51"}, "commit"=>"Save event", "group_path"=>"ga_pmi_dc_1"}
D, [2015-11-30T16:52:03.057583 #13092] DEBUG -- :   User Load (0.5ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = $1
LIMIT 1  [["id", 86]]
D, [2015-11-30T16:52:03.059951 #13092] DEBUG -- :   Group Load (0.2ms)  SELECT  "groups".* FROM "groups" WHERE "groups"."title
" = $1 LIMIT 1  [["title", "ga"]]
D, [2015-11-30T16:52:03.063080 #13092] DEBUG -- :   Group Load (0.2ms)  SELECT  "groups".* FROM "groups" WHERE "groups"."ances
try" = '1' AND "groups"."title" = $1 LIMIT 1  [["title", "pmi"]]
[...]

PR #287, opened 2015-12-03
Fixes Attendance bug #279

mattscilipoti:
Bug: PMI Instructor was unable to add Attendance events in the past.

- Moved decisions/validation to Model.  Controller looks like "normal" controller.
- Update test data (and associated specs) to support me logging in via github (locally).
- Added feature spec for marking attendance.  
  - ~~Unfortunately, it doesn't verify the last portion of attendance. :( I ran into a strange error while attempting to support javascript tests (required due to ajax calls).  The test_instructor we add for this spec disappears from db prematurely, with no evidence for the reason.  I will continue working on that issue (attendance_spec_with_javascript), but didn't want it to hold up this fix.~~
  - Solved.  Was configuration for transactional fixtures.


adambray:
Re: your issue with test data, I think I know the cause... you have to disable transactions in testing. it's because the JS driver runs 'out-of-process' and can't access the data.

adambray:
:+1:

spec/support/form_helpers.rb
```diff
@@ -0,0 +1,10 @@
+module FormHelpers
+  # Clicks the commit button regardless of id or text
+  def submit_form
```
adambray: Can we add a comment to document that this should be used in a `within` block?

bin/setup
```diff
@@ -11,6 +11,7 @@ Dir.chdir APP_ROOT do
   puts "== Installing dependencies =="
   system "gem install bundler --conservative"
   system "bundle check || bundle install"
+  system "brew install phantomjs" # for headless cayabara specs
```
adambray: Note here about phantomjs2 requirement, and issues with el capitan?

PR #292, opened 2016-01-01
CSS: Make buttons look more like buttons

RobertAKARobin:


adambray:
I played with it last night, and it looks fine to me!

adambray:
aka :+1:

mattscilipoti:
I appreciate your work on this.  I do feel that the buttons need some TLC.  The following comments are personal opinions, but they are based (I hope) on my experiences on the web. 
I haven't seen buttons like these in a while (rounded sides, obvious shadows, tiny text).  The web has been moving to a flat look, more subtle shading/shadows, subtle rounding, text that fills most of button.  The page you are reading this on is an example.  
I think we would be better served by changing the size to a standard button size.  I regularly confuse the current buttons as a horizontal rule, separating sections.  This is the first time I have seen buttons not sized to their content.
That said, I haven't been motivated enough to make the necessary changes.  You have.  By that alone, I feel that your opinion carries more weight.  I just wanted to throw it out to the group.

RobertAKARobin:
@mattscilipoti Yeah, shaded buttons are pretty "Web 1.0". Let me try it your way.

RobertAKARobin:
See: https://github.com/ga-dc/garnet/pull/294

PR #299, opened 2016-01-12
removed files that had no code in test file except for generated skeleton

andrewsunglaekim:


adambray:
Wondering if we should leave these as places to add tests soon...

PR #300, opened 2016-01-12
adds tags and taggings table

andrewsunglaekim:
adds `tag` and `taggings` model as well as migrations.
adds associations to `tags`, `taggings`, `memberships`.
adds test coverage for associations for `tags` and `memberships`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,7 @@ group :production do
 end
 
 group :development, :test do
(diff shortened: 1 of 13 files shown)


adambray:
This looks good, but has a merge conflict with the groups refactor branch. Propose we merge in the groups_refactor stuff first, then fix the conflicts.

PR #305, opened 2016-01-18
Improve cohort show performance

adambray:
This branch greatly improves the performance of the cohort show page, as well as greatly simplifies the related code.


adambray:
This should be applied after the memberships routes refactor.

PR #313, opened 2016-01-21
fixes issue where github info refreshes when student has deleted GH account

adambray:


adambray:
Andy gave the thumbs up, tested locally with no issues.

PR #315, opened 2016-01-22
Tagging ux

adambray:
This PR adds the ability to tag members with arbitrary tags in the cohort show page.

Tags are globally unique by name, but the UI is designed such that the tag list shows you all existing tags used in that cohort. If you want to add a new tag, just type in the name and hit enter. Members are entered using autocomplete as well.

These tags are now displayed in the membership roster and are searchable using jQuery DataTables plugin.


andrewsunglaekim:
:+1:

jshawl:
:+1: `create` method feels a little large, but I read through the logic twice and am not sure how this could be extracted to a model method

adambray:
Yeah, really can't come up with a way to do that better. I think there may be a way to do it in the model, but I'll mark a TODO and give that a shot later this week.

PR #316, opened 2016-01-22
Updates todo page to show only items in your 'squad'

adambray:
Does this by updating the get_due method to search for memberships with
tags in common with your admin memberships.

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -4,7 +4,7 @@ class DashboardsController < ApplicationController
 
     redirect_to current_user unless @user.is_admin_of_anything?
 
-    @due_submissions = @user.get_due("submissions")
-    @due_attendances = @user.get_due("attendances")
+    @due_submissions = @user.get_todo(Submission)
+    @due_attendances = @user.get_todo(Attendance)
   end
 end
--- a/app/models/attendance.rb
+++ b/app/models/attendance.rb
@@ -5,6 +5,11 @@ class Attendance < ActiveRecord::Base
 
   has_one :cohort, through: :event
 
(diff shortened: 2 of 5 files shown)


andrewsunglaekim:
Can we merge [tagging_ux](https://github.com/ga-dc/garnet/pull/315)? Would make it easier to see the diff for these PR's

adambray:
@andrewsunglaekim fine with that once that PR gets one more pair of :eyeglasses:

jshawl:
permission to rebase and force push to show the updated diff after https://github.com/ga-dc/garnet/commit/868367c1f028108595f429f29ab949b0ef6ecefe ?

adambray:
sure!

app/controllers/dashboards_controller.rb
```diff
@@ -4,7 +4,7 @@ def to_do
 
     redirect_to current_user unless @user.is_admin_of_anything?
 
-    @due_submissions = @user.get_due("submissions")
-    @due_attendances = @user.get_due("attendances")
+    @due_submissions = @user.get_todo(Submission)
+    @due_attendances = @user.get_todo(Attendance)
```
jshawl: have a feeling this should be a symbol a la `has_many` and `belongs_to`. is there any difference between doing something like this and `@user.get_todo :submission`

jshawl:
:+1: just comments for a future `TODO` or optimization but otherwise good to merge

app/controllers/dashboards_controller.rb
adambray: Keeps the actual method implementation (of `get_todo`) cleaner. No need to take the symbol and do `model.capitalize.constantize` etc.

PR #319, opened 2016-01-25
Let students take their own attendance

jshawl:


app/controllers/attendances_controller.rb
```diff
@@ -10,7 +10,16 @@ def create
 
   def update
     @attendance = Attendance.find(params[:id])
-    @attendance.update!(status: params[:status])
+    if params[:status]
+      @attendance.update!(status: params[:status])
+    else
+      # User is checking in
+      if request.remote_ip != ENV['check_in_static_ip']
```
adambray: I don't think this should be stored in an ENV var, or even used at all to enforce whether you can check in. At minimum, it should be inferred though `attendance.event.cohort.location`.

The thought behind using the event's cohort's location, is then it can be a config option per location.

The thought behind not enforcing IP (and using a 'password of the day' instead), is that some locations won't have a single static IP for each device on the network.

app/models/attendance.rb
```diff
@@ -22,4 +22,15 @@ def self.statuses
   def status_english
     return Attendance.statuses[self.status]
   end
+
+  def calculate_status
```
adambray: Few thoughts:

1. These are `magic numbers`, which [are bad](http://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad)
2. These numbers are too inflexible. What about classes that start at 10? Or 6:30pm? Instead, this should be based off the attendance's events's time.

app/views/memberships/show.html.erb
```diff
@@ -3,6 +3,14 @@
   <h2><%= link_to @user.name, @user %> in <%= link_to @membership.cohort.name, @membership.cohort %></h2>
 </div>
 
+<% if @attendances.last %>
```
adambray: We should probably find all attendances for events happening ± 3 hours from the current time instead. There's a chance we'll pre-create events for cohorts based on their regular schedule. (e.g. every Mon, Wed, at 6:30 from Jun 2nd to Aug 14th)

db/seeds/test_seed.rb
```diff
@@ -32,6 +32,8 @@
 pmi1.add_member(bob)
 pmi1.add_member(carol)
 
+Event.destroy_all
+Attendance.destroy_all
```
adambray: This should happen in the db/seeds script with the other destroy_all calls (in fact, they're already there)

app/views/memberships/show.html.erb
jshawl: this line avoids an error when an admin is looking at a students membership page, but agreed that attendances.last is wrong here

app/models/attendance.rb
jshawl: resolved in https://github.com/ga-dc/garnet/commit/3374788f511a13849d9b16aff92146582509f103

spec/models/attendance_spec.rb
```diff
@@ -1,5 +1,29 @@
 require 'rails_helper'
 
 RSpec.describe Attendance do
+  before(:all) do
+    load "#{Rails.root}/db/seeds/test_seed.rb"
+  end
+  describe "#calculate_status" do
 
+    before(:each) do
+      @evt = Event.create(title: "Test Event", date: Time.parse("1969-07-20 09:00:00"), cohort: Cohort.last)
```
jshawl: this is forcing capybara failure - https://travis-ci.org/ga-dc/garnet/builds/104702317 because there is a recent event, which hides the form. What's the right way to create and tear down after? cc @adambray

app/models/attendance.rb
adambray: :+1:

andrewsunglaekim: this is dope

readme.md
```diff
@@ -36,6 +36,7 @@ Garnet is a collaboration by the instructors of GA's WDI to produce a student da
     gh_client_id: "12345"
     gh_client_secret: "67890"
     gh_redirect_url: "http://localhost:3000/github/authenticate"
+    check_in_static_ip: "your.static.ip.address"
```
andrewsunglaekim: where is ip address being checked?

app/models/attendance.rb
mattscilipoti: Recommend ActiveRecord enum for the magic numbers for event status: http://api.rubyonrails.org/classes/ActiveRecord/Enum.html

app/views/memberships/show.html.erb
```diff
@@ -3,6 +3,15 @@
   <h2><%= link_to @user.name, @user %> in <%= link_to @membership.cohort.name, @membership.cohort %></h2>
 </div>
 
+<% if @is_current_user  %>
+  <% @current_attendances.each do |attendance| %>
+    <%= form_tag self_take_attendance_path(attendance), method: :put do  %>
```
mattscilipoti: Was there an issue with `form_for` and submit button?
```
form_for(attendance) do |f|
  f.submit...
```

app/models/attendance.rb
adambray: @mattscilipoti I totally agree (in fact I used AR::Enums for memberships status (see membership.rb).

I didn't want to modify that yet, but am creating an issue, and propose that we resolve this in a separate PR.

spec/models/attendance_spec.rb
adambray: seems to be working now, as the last commit builds successfully

readme.md
adambray: It's not, we need to remove this line. Will push a fix.

adambray: fixed in 227502e

readme.md
```diff
@@ -43,6 +43,9 @@ Garnet is a collaboration by the instructors of GA's WDI to produce a student da
 12. Log in with username `adam` and password `password`
 13. Enter the Konami code
 
+`check_in_static_ip` allows site administrators to restrict users "checking in" to a
```
adambray: remove this.

PR #322, opened 2016-01-25
updates ERD in readme

andrewsunglaekim:
Updates ERD to reflect current domain


jshawl:
can you use a gem to do this automatically? http://voormedia.github.io/rails-erd/

adambray:
:+1: for this PR.

I like the rails-erd tool, and would be down for including it as well.B ut at the same time, andy's ERD is so much nicer than the one generated by the app :)

PR #332, opened 2016-01-27
Add homework scores

beckybeauchamp1:
Needs Refactoring!

### DIFF
--- a/app/assets/javascripts/submission.js
+++ b/app/assets/javascripts/submission.js
@@ -1,5 +1,19 @@
 $(".submission-score").on("keydown", function(evt){
   if(evt.keyCode == 13){
-    console.log("this works!!!!");
+    var self = $(this);
+    var submissionId = $(this).attr("submission-id");
+    var url = "../submissions/" + submissionId + "/update_score";
+    $.ajax({
+      url: url,
+      dataType: 'json',
+      type: "put",
+      data: {
+        score: self.val()
+      }
+    }).done(function(res){
+      self.parent().parent().append($("<td>Score Update</td>"))
+    }).fail(function(res){
+      console.log("this failed");
+    });
   }
 });
--- a/app/assets/javascripts/submission.js
+++ b/app/assets/javascripts/submission.js
@@ -1,20 +1,18 @@
-$(".submission-score").on("keydown", function(evt){
-  if(evt.keyCode == 13){
-    var self = $(this);
-    var submissionId = $(this).attr("submission-id");
-    var url = "../submissions/" + submissionId;
-    $.ajax({
-      url: url,
-      dataType: 'json',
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/submission.js
```diff
@@ -0,0 +1,19 @@
+$(".submission-score").on("keydown", function(evt){
```
jshawl: feel like a `.on('submit')` is better / more flexible/semantic

app/controllers/submissions_controller.rb
```diff
@@ -23,6 +23,13 @@ def destroy
     redirect_to @submission.assignment
   end
 
+  def update_score
+    submission = Submission.find(params[:submission_id])
+    submission.score = params[:score]
+    submission.save
```
jshawl: `.update(score: params[:score])`

mattscilipoti: Why isn't this just a submission update, where params contains `{submission: { score: 2 }}`?  Do we need another action?

andrewsunglaekim: resolved, changed ajax to path to existing update, thanks @mattscilipoti !

app/assets/javascripts/submission.js
andrewsunglaekim: can only do `submit` on form tag, we only have input, should we just add a form wrapper for the input tag?

RobertAKARobin: Pretty sure "submit" is tricky and may vary a bit from browser to browser. I'd recommend sticking with `keydown`.

andrewsunglaekim: refactored to submit! was a little tricky, can always switch back to keydown if neccessary

app/assets/javascripts/submission.js
```diff
@@ -0,0 +1,18 @@
...
+  var url = "../submissions/" + submissionId;
+  $.ajax({
+    url: url,
+    dataType: 'json',
+    type: "put",
+    data: {
+      score: inputEl.val()
+    }
+  }).done(function(res){
+    inputEl.parents("tr").append($("<td>Score Update</td>"))
```
joe-gz: change parents to ".closest("tr")"

app/views/assignments/show.html.erb
```diff
@@ -34,6 +34,7 @@
       <% Submission.statuses.keys.each do |status| %>
         <td><%= status_button(submission, status) %></td>
       <% end %>
+      <td><form class="score-update-form"><input class="submission-score" submission-id="<%=submission.id%>" type="text" value="<%=submission.score%>"></form></td>
```
adambray: Can we make this `js-score-update-form`

PR #333, opened 2016-01-27
Added mantras

RobertAKARobin:


app/models/mantra.rb
```diff
@@ -0,0 +1,19 @@
+class Mantra
+  @@all = []
+
+  def self.reload
+    puts "Duck..." * 10
```
RobertAKARobin: Just to make this refreshing more visible in the server log

Issue #346, opened 2016-01-27
Width of submission score text box is too narrow on Safari


adambray:

andrewsunglaekim:
this isn't exclusive to safari. Happens in chrome too, when the `score update` td is appended makes the textbox only read 1 number

andrewsunglaekim:
fixed in [#349](https://github.com/ga-dc/garnet/pull/349)

Issue #343, opened 2016-01-27
implement fade out for hw score update message


amaseda:

andrewsunglaekim:
unnecessary due to updates in ui from [#349](https://github.com/ga-dc/garnet/pull/349)

PR #349, opened 2016-01-28
fixes #344 #346 #343

### LINKED ISSUE #344: Cleanup on HW Score
change parents to ".closest("tr")" in JS
use `js-` prefix for js classes.

andrewsunglaekim:
Removes appending td for score updates, instead flashes green when updating score
fixes overflow for tables for long usernames

### DIFF
--- a/app/assets/javascripts/submission.js
+++ b/app/assets/javascripts/submission.js
@@ -11,7 +11,9 @@ $(".js-score-update-form").on("submit", function(evt){
       score: inputEl.val()
     }
   }).done(function(res){
-    inputEl.parents("tr").append($("<td>Score Update</td>"))
+    var td = $("<td>Score Update</td>")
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/submission.js
```diff
@@ -11,7 +11,10 @@ $(".score-update-form").on("submit", function(evt){
       score: inputEl.val()
     }
   }).done(function(res){
-    inputEl.parents("tr").append($("<td>Score Update</td>"))
+    var td = $("<td>Score Update</td>")
```
adambray: is this `td` var/element used?

adambray: @andrewsunglaekim ^

andrewsunglaekim: updated and removed

PR #350, opened 2016-01-28
move cohort pill from layout to cohort show

amaseda:


app/views/cohorts/show.html.erb
```diff
@@ -1,3 +1,7 @@
+<% if @cohort %>
```
adambray: I don't think we need this `if` statement, since we're on a cohort show page.

adambray:
:+1:

PR #351, opened 2016-01-28
Event: rename column “date” to “occurred_at”

mattscilipoti:
- avoiding reserved words
- matches existing pattern: created/updated_at
- occurred_at should be *much* easier to search for. :)
- e.g. `if event.occurred_at.today?`.  Winner!


adambray:
TBH, I'm down with this, as it's a notable improvement, but I wonder if there isn't an even better name, since events can occur in the future? 

What about `occurs_on`?

adambray:
Also, thanks for making a PR that's strictly code cleanup!!!

mattscilipoti:
I feel occurred_on/occurs_on would indicate a date, not a datetime.

mattscilipoti:
The event occurred on Tuesday.  I agree, the "at" is a bit concise.  The occurred on Tuesday, at 2:00pm.  I could go with occurs_at.  Since it may be in the future.

mattscilipoti:
Making the change.

mattscilipoti:
@adambray Renamed to occurs_at.

adambray:
:+1:

andrewsunglaekim:
:+1:

PR #359, opened 2016-01-29
fix user, location and course associations

amaseda:


adambray:
@amaseda looks like you need to check that schema.rb was committed properly... the tests are failing because `schema.rb` doesn't match the migrations.

nolds9:
:+1:

jshawl:
travis failure: https://travis-ci.org/ga-dc/garnet/builds/105686293#L318

adambray:
:+1:

PR #364, opened 2016-01-29
Fixes #330 - improves user show view, tweaks cohort show view memberships table

### LINKED ISSUE #330: users show page should break memberships up
users show page should have separate tables for what their admins of vs what they are not admins of

joe-gz:
- removes if else logic on using cohort name vs user name
- changes table partial in the user show page to separate administrative cohorts v student cohorts

### DIFF
--- a/app/assets/javascripts/tables.js
+++ b/app/assets/javascripts/tables.js
@@ -2,8 +2,8 @@ $(".js-dataTable.memberships").DataTable({
     paging: false,
     "aoColumnDefs" : [
       {
-        "bSearchable": false,
-        "aTargets": [3, 4, 5, 6, 7]
+        "bSearchable": true,
+        "aTargets": [0,1]
       }
     ]
 });
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,10 @@ class ApplicationController < ActionController::Base
   before_action :authenticate
   helper_method :current_user, :signed_in?
 
+  rescue_from CanCan::AccessDenied do |exception|
+    redirect_to :back, :notice => exception.message
+  end
+
   rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
 
   private
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,6 +10,9 @@ class ApplicationController < ActionController::Base
   end
 
   rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
+  rescue_from CanCan::AccessDenied do |exception|
+     redirect_to root_url, :alert => exception.message
+  end
 
   private
     def record_not_found
--- a/app/controllers/cohorts_controller.rb
+++ b/app/controllers/cohorts_controller.rb
@@ -15,7 +15,7 @@ class CohortsController < ApplicationController
     @admins = @cohort.admins
 
     @assignments = @cohort.assignments.includes(:submissions)
-    @events = @cohort.events.includes(:attendances).order(date: :desc)
+    @events = @cohort.events.includes(:attendances).order(occurs_at: :desc)
 
     @event_for_today_already_exists = @events.on_date(Date.today).any?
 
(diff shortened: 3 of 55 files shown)


app/controllers/users_controller.rb
```diff
@@ -26,7 +26,12 @@ def show
     redirect_to current_user unless is_current_user || @is_adminned_by_current_user
 
     @is_editable = is_current_user && !@user.github_id
-    @memberships = @user.memberships.sort_by{|m| m.cohort.name}
+    @admin_cohorts = @user.cohorts_adminned_by(current_user)
+    @admin_memberships = @admin_cohorts.map do |cohort|
+      membership = Membership.find_by(cohort:cohort,user:@user)
+    end
+    @admin_memberships.sort_by{|m| m.cohort.name}
+    @non_admin_memberships = @user.memberships.sort_by{|m| m.cohort.name}-@admin_memberships
```
adambray: I would shorten this to:

```
memberships = @user.memberships.includes(:cohort).order("cohorts.name")
@admin_memberships = memberships.where(is_admin: true)
@student_memberships = memberships.where(is_admin: false)
```

adambray: In fact, with the right scopes on the Membership model. this could be improved even more:

```
memberships = @user.memberships.includes(:cohort).order("cohorts.name")
@admin_memberships = memberships.admin
@student_memberships = memberships.student
```

adambray:
I'd like to talk about this some more... big picture I have a few thoughts:

1. It's failing specs... want to find out why.
2. It adds a lot of code mass for not a major feature. I like where the UI is headed, but, why, for example, do we duplicate a lot of the memberships_table and _row  code just to customize it a bit (e.g. remove the observation column)?

andrewsunglaekim:
As per our discussion:

this PR removes both observations and image tag columns because what the student should see in a row differs from what an admin sees

Solution is either this, or if/else logic in the existing table to eliminate `th`s and `td`s for those that shouldn't exist from the student perspective.

app/assets/javascripts/tables.js
```diff
@@ -2,8 +2,8 @@ $(".js-dataTable.memberships").DataTable({
     paging: false,
     "aoColumnDefs" : [
       {
-        "bSearchable": false,
-        "aTargets": [3, 4, 5, 6, 7]
+        "bSearchable": true,
+        "aTargets": [0,1]
```
adambray: This is because the order of columns changed between tables.

app/controllers/application_controller.rb
```diff
@@ -10,6 +10,9 @@ class ApplicationController < ActionController::Base
   end
 
   rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
+  rescue_from CanCan::AccessDenied do |exception|
+     redirect_to root_url, :alert => exception.message
+  end
```
adambray: This needs to be removed, as it's duplicated above from another branch

app/controllers/cohorts_controller.rb
```diff
@@ -24,12 +27,12 @@ def show
...
-		  :disposition => "attachment; filename=#{@cohort.id}.csv"
-	else
-	  redirect_to @cohort, notice: "Requires admin rights to export"
-	end
+          send_data Cohort.to_csv(@student_memberships),
+          :type => 'text/csv; charset=UTF-8;',
+          :disposition => "attachment; filename=#{@cohort.id}.csv"
+        else
+          redirect_to @cohort, notice: "Requires admin rights to export"
+        end
```
adambray: correcting tabs -> spaces

app/models/membership.rb
```diff
@@ -45,7 +46,8 @@ def toggle_admin!
 
   def percent_from_status( association, status)
     assoc = self.send(association)
-    ((assoc.where(status:status).length.to_f / assoc.length.to_f) * 100).round(2)
+    return nil if assoc.length == 0
```
adambray: Can't calculate percentage if there's no divisor.

app/views/memberships/_row_performance_cells.html.erb
```diff
@@ -0,0 +1,14 @@
+<% data = [ membership.percent_from_status(:attendances, 2),
+            membership.percent_from_status(:attendances, 1),
+            membership.percent_from_status(:attendances, 0),
+            membership.percent_from_status(:submissions, 2) ] %>
+
```
adambray: Feel like this should be refactored somehow, not sure...

PR #367, opened 2016-01-29
fixes #265

### LINKED ISSUE #265: When admin looks at student view, include N/A
Shouldn't count in percentages, but should show up in rows.

joe-gz:
additionally allows student to view n/a which is different from original issue. Andy and I discussed that students would like to see future assignments, or past assignments that instructors may not have graded yet.


adambray:
:+1:

jshawl:
:+1:

PR #368, opened 2016-01-29
updates mantras to bring them into the app; fixes #361

### LINKED ISSUE #361: mantras should be self hosted?
What if @RobertAKARobin goes rogue? why not self-host with a mantra controller?

https://github.com/ga-dc/garnet/commit/ca2367ec24be235f05d6ac1a0eb733fadbbec927#diff-822b0d0191c78dd3ebae802cf2a2a2b8R8

jshawl:


jshawl:
:+1:

PR #373, opened 2016-01-29
remove orphans link

jshawl:


andrewsunglaekim:
was orphans a problem before? and why are they not now? is this something we're going to clean up in a different way going forward?

RobertAKARobin:
Orphans were users who had no memberships. I'm unsure whether this is still a problem.

adambray:
I don't think it's a problem, and would vote to remove this (i.e. merge this PR)

:+1:

RobertAKARobin:
:+1:

PR #375, opened 2016-01-30
hotfix: adds var to seed file

nolds9:


nolds9:
@jshawl @adambray  hotfix for seed file

PR #378, opened 2016-02-01
fixes #336

### LINKED ISSUE #336: background job to mark n/a attendances as missing one day later

andrewsunglaekim:
needs configuration to be run on server side still @adambray

### DIFF
--- a/app/models/attendance.rb
+++ b/app/models/attendance.rb
@@ -48,4 +48,9 @@ class Attendance < ActiveRecord::Base
       end
     end
   end
+
+  def self.mark_pastdue_attendances_as_missed
+    na_attendances = self.unmarked.joins(:event).where("events.occurs_at < ?", Time.now.beginning_of_day)
+    na_attendances.update_all(status: 0)
+  end
 end
--- a/app/models/attendance.rb
+++ b/app/models/attendance.rb
@@ -39,4 +39,13 @@ class Attendance < ActiveRecord::Base
       0
     end
   end
+
+  def mark_na_absent_24
+    unless self.status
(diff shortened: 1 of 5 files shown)


app/models/attendance.rb
```diff
@@ -39,4 +39,13 @@ def calculate_status
       0
     end
   end
+
+  def mark_na_absent_24
```
adambray: This should be a class method, and can we give it a more clear name?

Something like:

```
def self.mark_pastdue_attendances_as_missed
   #
end
```

Also, this method should use a where clause (along with a join on events) to only get past_due attendances from the DB.

As you have it currently, this involves fetching each record from the DB and checking it.

app/models/attendance.rb
```diff
@@ -39,4 +39,18 @@ def calculate_status
...
+    unless self.status
+      if Time.now > self.event.occurs_at + 24.hours
+        self.status = 0
+        self.save
+      end
+    end
+  end
```
adambray: plz remove

config/schedule.rb
```diff
@@ -0,0 +1,3 @@
+every :hour do
+  rake "attendance:mark_absent"
```
adambray: This should be every day at 6am

app/models/attendance.rb
```diff
@@ -39,4 +39,18 @@ def calculate_status
...
+
+  def self.mark_pastdue_attendances_as_missed
+    na_attendances = self.unmarked.joins(:event).where("events.occurs_at < ?", Time.now.beginning_of_day)
```
adambray: and actually, just realized due to time zones this should be (Time.now - 1.day)

PR #389, opened 2016-02-01
Auto add members to cohort

nolds9:


app/views/cohorts/show.html.erb
```diff
@@ -67,3 +67,11 @@
...
   </div>
 </section>
+
+<%= form_tag generate_invite_code_cohort_path(@cohort), method: "post" do  %>
+  <%= button_tag "Generate Invite Token", data: {confirm: "This will revoke all previous invite codes. Are you sure?"} %>
+<% end %>
+
+<% if @cohort.invite_code %>
+  <%= link_to "Invite Link", gh_authorize_url(invite_code: @cohort.invite_code) %>
+<% end %>
```
andrewsunglaekim: shouldn't these things be wrapped in some authorizations?

app/controllers/cohorts_controller.rb
```diff
@@ -82,6 +82,13 @@ def gh_refresh
     redirect_to cohort_path(@cohort)
   end
 
+  def generate_invite_code
+    @cohort = Cohort.find(params[:id])
```
adambray: @jshawl @nolds9 this could be set in a before filter for consistency

app/controllers/sessions_controller.rb
```diff
@@ -6,9 +6,9 @@ def new
     if current_user
       is_an_admin_of_anything = current_user.memberships.where(is_admin: true).length > 0
       if is_an_admin_of_anything
-	redirect_to root_path
+    	   redirect_to root_path
```
adambray: This indentation is now off.

app/models/user.rb
```diff
@@ -123,4 +124,12 @@ def is_member_of cohort
     cohort.memberships.exists?(user: self)
   end
 
+  private
+  def accept_invite
```
adambray: This feels like it should happen in a controller action... but I'm not really sure.

app/models/user.rb
```diff
@@ -123,4 +124,12 @@ def is_member_of cohort
...
+    if self.invite_code
+      cohort = Cohort.find_by(invite_code: self.invite_code)
+      cohort.memberships.create(user: self)
```
adambray: There's an interface method for this... `cohort.add_member(self)`

app/models/user.rb
jshawl: motivation was that two separate controllers handle accepting invites

adambray: word... I think that makes sense the more I think about it!

PR #392, opened 2016-02-01
jesse made me implement this feature

andrewsunglaekim:
need to update ui to look pretty, quick implementation so we can have at least the ability to delete members for now.


mattscilipoti:
Is there any authorization for delete?

PR #398, opened 2016-02-02
adjusts color in membership table to invert if value is negative

adambray:
e.g. tardiness

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -34,26 +34,26 @@ module ApplicationHelper
 
     case input
     when -1000...25
-      return "s0"
+      return "status_very_bad"
     when 25...50
-      return "s1"
+      return "status_bad"
     when 50...75
-      return "s2"
+      return "status_mediocre"
     when 75..1000
-      return "s3"
+      return "status_good"
     end
   end
 
   def color_of_status input
     case input * 100
     when 0...50
-      return "s0"
+      return "status_very_bad"
     when 50...100
-      return "s1"
+      return "status_bad"
     when 100...150
-      return "s2"
+      return "status_mediocre"
     when 150..200
-      return "s3"
+      return "status_good"
     end
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -26,6 +26,9 @@ module ApplicationHelper
   end
 
   def color_of_percent input, type = :good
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -25,29 +25,38 @@ def average_status collection
...
+    # if the input is a 'bad' value, flip and scale the value accordingly.
+    # e.g. a good value would be %HW complete, and 75% would be green
+    #
+    # but a bad value might be "% absent", and 10% would be yellow
+    input = (100 - (input* 3)) if type == :bad
+
     case input
-    when 0...25
-      return "s0"
+    when -1000...25
```
jshawl: is it sometimes negative? is this magic?

app/helpers/application_helper.rb
```diff
@@ -25,29 +25,38 @@ def average_status collection
...
     when 25...50
-      return "s1"
+      return "status_bad"
     when 50...75
-      return "s2"
-    when 75..100
-      return "s3"
+      return "status_mediocre"
+    when 75..1000
+      return "status_good"
```
jshawl: printing CSS classes from a ruby method feels like a code smell. why the change from integers? also maybe a data-attribute is the best fit here:

```html
<p data-status="<%= @status %>>"</p>
```

app/helpers/application_helper.rb
adambray: Agree that the numbers should be extracted into named constants, good catch.

As for the values, they can be negative in the case where the input is of type `:bad`. See the formula and how it subtracts a multiple of input from 100. So a 'bad' input of 100 would be (100 - (3*100) == -200.

This sort of formula is necessary to scale bad inputs appropriately. Probably an opportunity for refactoring here though.

app/helpers/application_helper.rb
adambray: Printing CSS classes using ERB is a pretty common pattern that I've seen... and in this case, this change isn't introducing that... it was already printing CSS classes... this PR just renames the classes to be more clear.

Using `data-status` would require JS just to apply the styles, which would introduce additional complexity and indirection.

adambray: To be clear, these weren't integers before, they were CSS classes, eg `"s0"`.

adambray:
@jshawl followup thoughts?

jshawl:
:+1:

PR #400, opened 2016-02-02
updates manage page to include links for activate / deactivate, and...

adambray:
fixes corresponding bug in the memberships table caching

### DIFF
--- a/app/views/cohorts/show.html.erb
+++ b/app/views/cohorts/show.html.erb
@@ -9,14 +9,17 @@
   <div>
     <h4>Roster</h4>
 
-    <% cache cache_key_for_cohort(@cohort, @active_memberships, current_user) do %>
+    <%  # these two tables are both cached using @cohort.memberships rather than
+        # their corresponding subset to handle the edge case where a user is
+        # activated or de-activated %>
+    <% cache cache_key_for_cohort(@cohort, @cohort.memberships, current_user, "active") do %>
       <%= render partial: "memberships/table", locals: {
                                                 memberships: @active_memberships,
                                                 header: "Active Members",
                                                 row_header: :user_name } %>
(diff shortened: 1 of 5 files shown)


app/views/cohorts/show.html.erb
```diff
@@ -9,14 +9,17 @@
   <div>
     <h4>Roster</h4>
 
-    <% cache cache_key_for_cohort(@cohort, @active_memberships, current_user) do %>
+    <%  # these two tables are both cached using @cohort.memberships rather than
+        # their corresponding subset to handle the edge case where a user is
+        # activated or de-activated %>
+    <% cache cache_key_for_cohort(@cohort, @cohort.memberships, current_user, "active") do %>
```
jshawl: mismatch erb tags from multiline comment?

app/views/cohorts/show.html.erb
```diff
@@ -9,14 +9,17 @@
   <div>
     <h4>Roster</h4>
 
-    <% cache cache_key_for_cohort(@cohort, @active_memberships, current_user) do %>
+    <% # these two tables are both cached using @cohort.memberships rather than %>
+    <% # their corresponding subset to handle the edge case where a user is %>
+    <% # activated or de-activated %>
```
adambray: I'm ok with these changes, but were they strictly necessary? I tested this and the app worked fine for me. In other words, I don't think the ERB tags were mismatched, just multiline, which is allowed in ERB, AFAIK.

jshawl: not necessary - just a fix for syntax highlighting

jshawl:
:+1: good 2 merge

andrewsunglaekim:
I like the table alot better  =D :+1:

PR #401, opened 2016-02-02
adds view + specs for member's average observations

nolds9:
@andrewsunglaekim


app/views/memberships/_row.html.erb
```diff
@@ -8,6 +8,6 @@
     <td><%= tag_list(membership)%></td>
     <%= render partial: "memberships/row_performance_cells", locals: {membership: membership} %>
     <% end %>
-
```
nolds9: remove space?

andrewsunglaekim:
this looks great and the testing! But I remember the code base had something like this before? Why was it taken out in the first place? I could be totally mistaken, But either way :+1:

RobertAKARobin:
:+1:

PR #407, opened 2016-02-02
fixes bug where anyone could destroy a membership

adambray:


jshawl:
:+1:

andrewsunglaekim:
:+1:

app/controllers/memberships_controller.rb
```diff
@@ -16,13 +16,13 @@ def create
   end
 
   def show
+    authorize! :show, @membership
```
nolds9: shouldn't this be `:read`?

adambray: I think you're right.

-Adam

adambray: @nolds9 you were right that `:read` is the traditional name of this permission, so I changed it (it was actually used in 3 places).

Of note, I learned that `:show` works as well because CanCanCan alias show to read (and a few more!) https://github.com/CanCanCommunity/cancancan/wiki/Action-Aliases

PR #409, opened 2016-02-02
Fixes bug where user shown as not registered if they have capital let…

jshawl:
…ters in their name

https://github.com/ga-dc/garnet/blob/06cdce7802273c4b095f6b042af9cc680088ebfc/app/models/user.rb#L23-L31


nolds9:
:thumbs_up

RobertAKARobin:
:+1:

PR #412, opened 2016-02-03
fixes #411, average observations as NaN

### LINKED ISSUE #411: Average Observations show NAN
To reproduce: view http://garnet.wdidc.org/cohorts/4#members

then view http://garnet.wdidc.org/memberships/499

nolds9:


spec/models/membership_spec.rb
```diff
@@ -25,6 +25,20 @@
     it "has an average observations method" do
       expect(@m.average_observations).to eq(0.33)
     end
+    it "has average observations that are a float" do
+      expect(@m.average_observations).to be_a(Float)
+    end
+    it "has average observations equal to zero as a float " do
```
nolds9: would like to improve this wording

spec/models/membership_spec.rb
```diff
@@ -25,6 +25,20 @@
...
+      cohort = Cohort.first
+      @t = cohort.memberships.create()
+      @t.observations.create(status: 0)
```
nolds9: not sure where to move this code, was trying to test edge cases

app/models/membership.rb
```diff
@@ -51,7 +51,7 @@ def percent_from_status( association, status)
   end
 
   def average_observations
-    (self.observations.map(&:status).inject(:+).to_f/(self.observations.length)).round(2)
+    !self.observations.empty? ? (self.observations.map(&:status).inject(:+).to_f/(self.observations.length)).round(2) : 0.0
   end
```
andrewsunglaekim: might be picky, but can we abstract `self.observations.map(&:status).inject(:+).to_f/(self.observations.length)` into a variable? into

andrewsunglaekim:
:+1:

app/models/membership.rb
```diff
@@ -51,7 +51,7 @@ def percent_from_status( association, status)
...
 
   def average_observations
-    (self.observations.map(&:status).inject(:+).to_f/(self.observations.length)).round(2)
+    !self.observations.empty? ? (self.observations.map(&:status).inject(:+).to_f/(self.observations.length)).round(2) : 0.0
```
andrewsunglaekim: Talked with matt about this, and its probably grey area. But since its relatively complex code inside of ternary, might be nice to abstract `self.observations.map(&:status).inject(:+).to_f/(self.observations.length)` into an intermediary variable.

adambray: First comment... should this return 0.0 if there are no observations?

On the one hand, I like keeping return value consistent, but on the other hand, 0.0 isn't technically correct either... to me this scenario might call for returning `nil`.

RobertAKARobin: Yeah, this is a little gnarly. I'd like to see this made more readable, too.

spec/models/membership_spec.rb
RobertAKARobin: How about `it "'s average observation score is 0.0" do`?

app/models/membership.rb
adambray: Second... assuming we keep it returning 0, I'd vote to do to do a few things:

1. convert the double negation in the condition to a positive.
2. extract into a variable as you mentioned in the comments:

```
def average_observations
  average = (self.observations.map(&:status).inject(:+).to_f/(self.observations.length)).round(2)
  self.observations.any? ? average : 0.0
end
```

adambray: c.f. this code with the method directly above it, which returns nil if there are no associations:
https://github.com/ga-dc/garnet/blob/master/app/models/membership.rb#L47

adambray: Correspondingly, the view that uses the similar method handles nil values:
https://github.com/ga-dc/garnet/blob/master/app/views/memberships/_row_performance_cells.html.erb#L8

A change to this method would require the same if statement for observations here:
https://github.com/ga-dc/garnet/blob/master/app/views/memberships/_row_performance_cells.html.erb#L8

adambray:
:+1:

jshawl:
@adambray failing travis build? https://travis-ci.org/ga-dc/garnet/builds/107000873

PR #413, opened 2016-02-03
Adds Membership API and Token-Based Auth for users

nolds9:


app/controllers/cohorts_controller.rb
```diff
@@ -90,6 +90,13 @@ def generate_invite_code
     redirect_to :back
   end
 
+  def generate_api_token
+    @cohort = Cohort.find(params[:id])
```
adambray: This is best done in the before_action

app/views/api/cohorts/get_api_token.html.erb
```diff
@@ -0,0 +1,11 @@
+<% if current_user && current_user.is_member_of(@cohort) %>
```
adambray: What is the purpose of this method? If this works as I understand, it's dangerous for two reasons:

1) it gives access to API tokens to any member of a cohort
2) it allows the API token to be grabbed by a malicious 3rd party site via CSRF attacks (because we've enabled CORS)

config/application.rb
```diff
@@ -24,5 +24,11 @@ class Application < Rails::Application
 
     # Do not swallow errors in after_commit/after_rollback callbacks.
     config.active_record.raise_in_transactional_callbacks = true
+    config.middleware.insert_before 0, "Rack::Cors" do
```
adambray: :warning: this seems really dangerous, as it opens the app up to CSRF attacks on all the data exposed through the API, including membership data. I would suggest we keep CORS disabled until we have a clear idea of how to safely enable it, or at least only whitelist wdidc.org and localhost as origins.

In the meantime, any apps we build can use a simple backend to fetch data.

adambray:
One other thought... any reason we're granting API tokens at the cohort level? My feeling is that tokens on a per user basis is better in a few ways (which I thought we talked about and agreed upon?):

1. we can audit who's doing what if need be
2. a user can reject / replace their own personal auth token without impacting others
3. simpler to code and maintain as the APIs would have access to `current_user` and thus can take advantage of all of the user-based cancancan auth code.
3a. without per-user API auth, we can't serve any data that is sensitive per user, such as observations
4. related to #3 above, the APIs could be designed to be more restful (using IDs for cohort lookup) and allowing 1 token to be used to fetch all data I can see.

jshawl:
User specific api tokens is totally the right move here. Wanna pair on this? Maybe after install fest?

jshawl:
to reproduce flickering test: 

`rspec` errors

`rspec spec/features/attendance_spec.rb` doesn't

app/models/ability.rb
```diff
@@ -1,7 +1,8 @@
 class Ability
   include CanCan::Ability
 
-  def initialize(user)
+  def initialize(user, params)
+    user ||= User.find_by(api_token: params[:api_token])
```
adambray: @jshawl filed an issue to refactor this, so I'm ok with merging it for now.

adambray:
:+1:

PR #415, opened 2016-02-04
Fixes #335: Updates Attendance/Submissions to use AR enums

amaseda:
Replace attendance/submission statuses class method with enums

PR #416, opened 2016-02-04
changes observations statuses class method  into enum

andrewsunglaekim:
kind of a follow on to [PR #415](https://github.com/ga-dc/garnet/pull/415)


RobertAKARobin:
Cool! Didn't know this existed. :+1:

app/views/observations/_observation.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="observation <%= color_of_status(observation.status) %>">
+<div class="observation <%= color_of_status(observation.get_enum_status_int) %>">
```
adambray: just a thought... do we need this new method? `observation.status` will now return `:red` for example, which I think will get stringified as `red` into the HTML... so what if we updated the css so that the class names weren't numeric but red, yellow and green?

Then everything is more semantic!

app/models/membership.rb
```diff
@@ -51,7 +51,8 @@ def percent_from_status( association, status)
   end
 
   def average_observations
-    average = (self.observations.map(&:status).inject(:+).to_f/(self.observations.length)).round(2)
+    status_integers = self.observations.map{|observation| observation.get_enum_status_int}
+    average = (status_integers.inject(:+).to_f/(self.observations.length)).round(2)
```
adambray: I just learned something new today! ActiveRecord has a pre-included [calculations API](http://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-average)!!!

Try this:
```
average = self.observations.average(:status)
self.observations.any? ? average : nil
```

andrewsunglaekim:
Latest commit, updates based off adam's comments

@adambray ^ probably good to merge with attendance-enums

adambray:
:+1:

PR #420, opened 2016-02-05
remove pending tests

jshawl:


RobertAKARobin:
:+1:

nolds9:
:+1:

andrewsunglaekim:
i think i had tried to remove some of these in a prior PR, but Adam recommended against it. As it kind of acts like a todo for testing? I dunno, would like adam's thoughts on this PR. @adambray

adambray:
@jshawl @andrewsunglaekim the more I look at these, the more I'm ok with removing them. A lot has changed, and most of the group stuff isn't relevant anymore.

Probably favor starting with acceptance testing most things, and doing so from scratch (excluding existing acceptance tests that work).

PR #423, opened 2016-02-08
adds pie charts to membership show page for data vis purposes

andrewsunglaekim:
This needs to badly be refactored, and my first time using high charts. But i think it's a good first step. Needs a lot of work.


app/controllers/memberships_controller.rb
```diff
@@ -54,6 +54,23 @@ def toggle_admin
     redirect_to :back
   end
 
+  def student_report
```
jshawl: move this method to api controller - https://github.com/ga-dc/garnet/blob/1c6a4ff9258e3c0a4ab6f491f9642326284ae7c7/app/controllers/api/memberships_controller.rb

app/controllers/memberships_controller.rb
```diff
@@ -54,6 +54,23 @@ def toggle_admin
...
+        incompletes: @submissions.incompletes.length,
+        completes: @submissions.completes.length,
+      },
+      attendances: {
+        absences: @attendances.absences.length,
+        tardies: @attendances.tardies.length,
+        presences: @attendances.presences.length,
+      }
+    }
+    render json: report
```
jshawl: should use cancancan ability for authorization

PR #424, opened 2016-02-08
Fixes issue where user could not view any membership page

jshawl:
:fire_engine:   steps to reproduce: view a membership show page while not an admin.


jshawl:
Prevents users from self attendance cc @adambray @andrewsunglaekim

RobertAKARobin:
:+1:

adambray:
:+1:

PR #428, opened 2016-02-08
nest assignments under cohort

jshawl:
easier to move back wip please comment


adambray:
@jshawl could you clarify the reasoning behind this? I'm totally down with it technically, just want to understand the why!

jshawl:
navigability via the url

adambray:
Do we need to navigate by URL? If that's the only reason, could we improve the UX / views to support easier navigation?

jshawl:
closing - stale

PR #430, opened 2016-02-08
fixes observations so they only show on the correct page

### LINKED ISSUE #429: Observations should be specific to membership
https://github.com/ga-dc/garnet/blob/4a38d4062460a59b17f5af1113aa4a088516f938/app/controllers/memberships_controller.rb#L38

adambray:
Fixes #429


adambray:
`records_accessible` is no longer used or needed, and thus I removed it as well.

nolds9:
:+1:

jshawl:
:+1:

PR #432, opened 2016-02-09
removes average_status method from application helper

andrewsunglaekim:
If anyone was keeping this around for any reason..speak now or forever hold your peace.


jshawl:
:+1: save merge conflicts. Also, grep the code base for usage?

andrewsunglaekim:
yep, only place it existed was in the application helper

adambray:
:+1

PR #435, opened 2016-02-10
exclude future attendances from percentage present calculation

jshawl:


nolds9:
:+1:

app/models/membership.rb
```diff
@@ -46,6 +46,7 @@ def toggle_admin!
 
   def percent_from_status( association, status)
     assoc = self.send(association)
+    assoc = assoc.joins(:event).where("events.occurs_at < ? ", Time.now) if(association == :attendances)
```
adambray: @jshawl check out the [existing named scope](https://github.com/ga-dc/garnet/blob/master/app/models/attendance.rb#L9) for `due`!

I suspect you could use it here?

```
assoc = assoc.due if association == :attendances
```

spec/models/membership_spec.rb
```diff
@@ -8,6 +8,9 @@
     @m.observations.create(status: 0)
     @m.observations.create(status: 0)
     @m.observations.create(status: 1)
+    cohort.events.create!(title:"Some test event2", occurs_at: Time.now)
+    @m.attendances.first.update(status:2)
```
adambray: You can use AR enum bang methods here after the other PR (#415) gets merged in:

```
@m.attendances.first.present!
```

app/models/membership.rb
adambray: Related: My new years resolution is to never write anything more than a simple `where` clause on a model class outside of it's definition file.

jshawl:
@adambray can haz merge? this is blocking pre-creating attendance events

adambray:
:+1:

PR #438, opened 2016-02-10
clicking on a tag filters by that tag

jshawl:


app/assets/javascripts/tables.js
```diff
@@ -1,4 +1,4 @@
-$(".js-dataTable.memberships").DataTable({
+var membershipsDataTable = $(".js-dataTable.memberships").DataTable({
```
jshawl: global :(

app/assets/javascripts/tags.js
```diff
@@ -14,3 +16,9 @@ function removeTag(event) {
       el.closest("li").fadeOut();
     });
 }
+
+function filterByTag(event){
+  if(event.target.classList.contains("js-tag")){
```
adambray: use jQuery?

untested:
```
var target = $(event.target);
if (target.hasClass('js-tag')) {
...
```

adambray:
:+1: The suggestion I made isn't necessary, so support merging in either way.

PR #444, opened 2016-02-11
Cleans up API Token Auth Code

adambray:
- ensures a blank (empty string) token won't be checked against DB
- ensures all new users get an API token
- provides a rake task to auto generate api tokens for all existing users


app/controllers/application_controller.rb
```diff
@@ -31,7 +31,7 @@ def set_current_user model
 
     def current_user
       begin
-        if params[:api_token]
+        if params[:api_token].present?
```
jshawl: Can you write a test showing this bug in action

lib/tasks/api_tokens.rake
```diff
@@ -0,0 +1,7 @@
+namespace :api do
+  desc "Generates API tokens for all users which don't have one"
```
jshawl: Why rake task instead of migration?

app/controllers/application_controller.rb
adambray: @jshawl done!

lib/tasks/api_tokens.rake
adambray: @jshawl I'm honestly on the fence about this... I'm starting to feel like small things like this are better done in a rake to keep migrations clean.

adambray: Also, it'll encourage us to come up with a good way to run rake tasks on wdidc.org

app/models/user.rb
```diff
@@ -123,10 +125,6 @@ def as_json(options={})
     super.as_json(except: :password_digest)
   end
 
-  def generate_api_token
-    update(api_token: Digest::MD5.hexdigest(self.name + Time.now.to_s))
```
adambray: @jshawl I believe this was insecure btw... knowing a user's name, you could just iterate over all reasonable values of the time their token was generated, and keep trying until you find it.

jshawl: oh wow good point.. oops

lib/tasks/api_tokens.rake
jshawl: `garnet run ...` ?

jshawl:
:+1:

nolds9:
:+1:

PR #449, opened 2016-02-12
Exclude na from percentage calculation

jshawl:


RobertAKARobin:
:+1:

spec/models/membership_spec.rb
```diff
@@ -11,6 +11,10 @@
     cohort.events.create!(title:"Some test event2", occurs_at: Time.now)
     @m.attendances.first.update(status:2)
     cohort.events.create!(title:"Some test event", occurs_at: Time.now + 3.hours)
+    cohort.assignments.create!(title: "Some test assignment")
+    cohort.assignments.create!(title: "Some test assignment #2")
+    @m.submissions.first.update(status: 2)
```
adambray: Can we use the enum methods for clarity here?

e.g.:

```ruby
Submissions.statuses[:completed]
# or
Submissions.statuses[:or]
```

spec/models/membership_spec.rb
```diff
@@ -48,5 +52,8 @@
     it "should not include future attendances" do
       expect(@m.percent_from_status(:attendances, 2)).to eq(100)
     end
+    it "should exclude n/a assignments" do
+      expect(@m.percent_from_status(:submissions, 2)).to eq(100)
```
adambray: Same here (using enums)

app/models/membership.rb
```diff
@@ -48,7 +48,7 @@ def percent_from_status( association, status)
     assoc = self.send(association)
     assoc = assoc.due if(association == :attendances)
     return nil if assoc.length == 0
-    ((assoc.where(status:status).length.to_f / assoc.length.to_f) * 100).round(0)
+    ((assoc.where(status:status).length.to_f / assoc.where.not(status:nil).length.to_f) * 100).round(0)
```
adambray: n/a is now represented by a `0` due to AR::enums, and is called `unmarked`. I think this works better:

```
assoc.where.not(status: :unmarked)
``

File: spec/models/membership_spec.rb (hunk 1 above)
Comment (adambray) [inline spec/models/membership_spec.rb]: Actually, even better is to use the 'bang' methods:

```ruby
@m.submissions.first.completed!
```

app/models/membership.rb
jshawl: this one fails

adambray: You should at least be able to do `assoc.where.not(status: assoc.statuses[:unmarked])`

adambray:
:+1:

PR #454, opened 2016-02-12
memberships api filter by tag

andrewsunglaekim:


jshawl:
:+1:

app/controllers/api/memberships_controller.rb
```diff
@@ -4,13 +4,16 @@ class MembershipsController < ApplicationController
     def index
       @cohort = Cohort.find(params[:cohort_id])
       if can? :list_cohort_members, @cohort
-        @names = @cohort.memberships.map do |member|
-          member.user.name
+        @members = @cohort.memberships
+        if params[:tag]
+          @members = @members.filter_by_tag(params[:tag])
         end
+        @names = @members.map{|member| member.user.name}
```
adambray: should we render memberships, and include names?

app/models/membership.rb
```diff
@@ -56,4 +56,9 @@ def average_observations
     self.observations.any? ? average : nil
   end
 
+  def self.filter_by_tag tag
```
adambray: How is this working as a `self` (aka class) method?

adambray: ohhhh, it's on a result set.

adambray:
:+1:

PR #461, opened 2016-02-12
fixes bug where NaN generated if no marked submissions/attendances are present

adambray:
Will file a test for this shortly, but this could crash in production.


RobertAKARobin:
:+1:

andrewsunglaekim:
:+1:

PR #464, opened 2016-02-12
changes raise to a flash notice and redirects to root path

andrewsunglaekim:
old code from when we had a global rescuer


RobertAKARobin:
:+1:

adambray:
:+1:

PR #475, opened 2016-02-14
Removed jQuery UI

RobertAKARobin:
Adds 59 stylesheets and scripts just to make submissions flash when a score is entered. Changed to use existing "waiting" class instead.


jshawl:
:+1:

RobertAKARobin:
Thumb from Adrian

Issue #465, opened 2016-02-12
skip travis option


jshawl:
options:

test branch name?

RobertAKARobin:
Why's Travis sooo slooooow?

jshawl:
you can skip travis like so - https://docs.travis-ci.com/user/customizing-the-build/#Skipping-a-build

To deploy, you'll need commit rights to `git@garnet.wdidc.org:garnet.git`

PR #477, opened 2016-02-15
add skip option

jshawl:


jshawl:
fixes #465

jshawl:
this still has to wait for the build to start, so technically the only thing being skipped is running through the test suite.

Alternatively, we can designate select users who can push to `git@garnet.wdidc.org`

PR #479, opened 2016-02-16
github issues link by tag; re #478

jshawl:


app/models/cohort.rb
```diff
@@ -49,6 +49,10 @@ def add_member(user, is_admin = false)
     self.memberships.create!(user: user, is_admin: is_admin)
   end
 
+  def tags
+    self.memberships.map { |m| m.tags }.flatten.uniq
+  end
```
jshawl: can this be optimized w/ sql?

app/views/assignments/show.html.erb
```diff
@@ -13,6 +13,17 @@
 <% status_link_text = @show_na ? "Show All" : "Show only N/A" %>
 <% active_link_text = @show_inactive ? "Show only active students" : "Include inactive students" %>
 
+<div>
+  Issues by tag:
+  <ul>
+  <% @assignment.cohort.tags.each do |tag| %>
+    <li>
+      <a href="<%= @assignment.issues_url(tag.memberships.map{|m| m.user }) %>"><%= tag.name %></a>
```
jshawl: gnarly. should probably be a helper.

or maybe a method on tag? e.g. tag.users

or has_many tags
has many users through tags

PR #481, opened 2016-02-16
Fix 453

jshawl:


jshawl:
https://travis-ci.org/ga-dc/garnet/builds/109722374#L338-L343

PR #482, opened 2016-02-17
Added membership averaging methods

RobertAKARobin:
- An alternative to using `cache` because it's kinda scary, and some data appears to not be refreshing
- percent_from_status takes status name, not number

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -6,6 +6,10 @@ class Assignment < ActiveRecord::Base
 
   after_create :create_submissions
   after_initialize :set_defaults
+  after_update do
+    # Trigger recalculating of averages
+    self.submissions.each{ |s| s.save }
+  end
 
   attr_accessor :issues
 
--- a/app/models/attendance.rb
+++ b/app/models/attendance.rb
@@ -12,6 +12,10 @@ class Attendance < ActiveRecord::Base
 
   enum status: [:unmarked, :absent, :tardy, :present]
 
+  after_save do
+    self.membership.update_percents_of("attendance")
+  end
+
   def date
     self.event.occurs_at.strftime("%a, %m/%d/%y")
   end
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -1,4 +1,5 @@
 class Membership < ActiveRecord::Base
+  include ModelHelpers
   enum status: [ :active, :inactive ]
 
   belongs_to :cohort
@@ -12,7 +13,9 @@ class Membership < ActiveRecord::Base
 
   has_many :observations
   has_many :attendances
+  serialize :percent_attendances, JSON
   has_many :submissions
+  serialize :percent_submissions, JSON
 
   validate :is_unique_in_cohort, on: :create
   before_save :convert_nil_to_false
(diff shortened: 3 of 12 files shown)


app/models/attendance.rb
```diff
@@ -32,4 +34,9 @@ def self.mark_pastdue_attendances_as_missed
     na_attendances = self.unmarked.joins(:event).where("events.occurs_at < ?", Time.now.beginning_of_day)
     na_attendances.update_all(status: Attendance.statuses[:absent])
   end
+
+  private
+  def update_member_averages
+    self.membership.update_percentages_of("attendances")
```
jshawl: Can you write a test for this model method?

RobertAKARobin:
Will add tests soon!

app/views/memberships/_row.html.erb
```diff
@@ -1,19 +1,17 @@
-<% cache [membership, (can? :see_performance, membership), (can? :see_observations, membership)] do %>
-  <tr>
-    <% user = membership.user %>
-    <td><%= link_to user.name, membership %></td>
-    <td><%= avatar(user) if user.image_url && !user.image_url.strip.blank? %></td>
+<tr>
```
RobertAKARobin: Just removed caching and unindented

RobertAKARobin:
I think this homeboy's ready to roll.

app/models/assignment.rb
```diff
@@ -6,6 +6,10 @@ class Assignment < ActiveRecord::Base
 
   after_create :create_submissions
   after_initialize :set_defaults
+  after_update do
+    # Trigger recalculating of averages
+    self.submissions.each{ |s| s.save }
+  end
```
adambray: instead of calling save, is it possible to write a method that triggers just the calculation of averages?

adambray:
I'm can't give this a full review today. Can we hold off on merging this in, and I'll spend some time tonight reviewing it?

app/models/assignment.rb
RobertAKARobin: I like `save` because there may be other things we want to happen when an event is updated, and they can all be triggered with `after_save`. Also, if the name of the method changes it would need to be changed in each other model.

amaseda:
:+1:

Could go either way with the save vs. method thing tbh.

app/models/assignment.rb
adambray: Ok, I'm ok with that.

app/models/membership.rb
```diff
@@ -45,18 +47,41 @@ def toggle_admin!
...
+    self.update!(:"percent_#{plural}" => percents)
+    return percents
   end
 
   def average_observations
+    average = self.read_attribute("average_observations")
+    if !average
+      average = update_average_observations
+    end
+    return average
```
RobertAKARobin:
```rb
return self.read_attribute("average_observations") || update_average_observations
```

adambray:
Ok, so my general feeling is that this isn't a bad idea, but to be honest, this is a +104 lines of code, and minus ~25 (when you factor in comments removed)... and more importantly to me, I don't really understand what's going on in the `update_percent_of` method...

adambray:
In other words, I'm worried about increasing the complexity of the app here.

Admittedly the caching wasn't simple, but most of that complexity was limited to 2 lines, and it was commented well. This spreads the complexity all over the place, and isn't commented. @RobertAKARobin  can you at least add some comments and I'll take another read.

RobertAKARobin:
I think it's pick your poison because I don't really understand what's going on with `cache`.

I think `update_percents_of` is semantic when you see it in action:

```
@membership.update_percents_of "submissions"
```

app/models/membership.rb
```diff
@@ -45,18 +47,41 @@ def toggle_admin!
...
+    end
+    plural = "#{model.name.downcase}s"
+    relations = self.send(plural).due
+    num_total = relations.where.not(status: model.statuses[:unmarked]).length
+    model.statuses.each do |status, status_num|
+      percent = num_total < 1 ? 0 : (relations.where(status: status_num).length / num_total.to_f)
+      percents[status] = (percent * 100).round(0)
+    end
+    self.update!(:"percent_#{plural}" => percents)
+    return percents
```
adambray: This part in particular is tricky.

RobertAKARobin:
@adambray I've updated with comments.

adambray:
@RobertAKARobin I'm feeling ok about the code, but one thing I noticed is that removing the caching  seems to impact performance in the cohort show page pretty dramatically (from less than .5 sec to 3+ seconds).

This is because most of the cost of views is in rendering the partials and (usually to a lesser degree, including here) making all the DB calls. Caching eliminates all of that.

Any way to to do both here? Cache the values in the DB and the views?

RobertAKARobin:
@adambray Fair point. I've put the caching back in, so now this PR just adds the averaging methods and leaves the cache stuff alone. Best of both worlds?

jshawl:
:+1:

adambray:
:+1:

adambray:
Can we wait to merge this in until EOD in case we have any issues?

RobertAKARobin:
I'ma close this since it doesn't really do anything new. May revisit in the future.

RobertAKARobin:
It's baaaack. I think this will fix https://github.com/ga-dc/garnet/issues/520

RobertAKARobin:
This got thumbs and hasn't changed since it was last opened, so I'ma merge it.

PR #488, opened 2016-02-20
Custom sorting

RobertAKARobin:
Added my own table sorter instead of using the dataTables jQuery extension. Advantage is it gives us greater control over the styling and can add custom filters as need-be.


app/assets/javascripts/helpers.js
```diff
@@ -0,0 +1,50 @@
+"use strict";
+
+var h = (function(){
+  return {
+    el: el,
+    forEach: forEach,
+    is_a: is_a,
+    collect: collect
```
jshawl: maybe underscore or lodash is a better choice here?

app/assets/javascripts/tables.js
```diff
@@ -1,19 +1,134 @@
...
+      var rx = new RegExp(value, "i");
+      if(rx.test(row.text)) row.el.removeAttribute("data-hidden");
+      else row.el.setAttribute("data-hidden", true);
+    });
+  }
+  function filterOff(){
+    var instance = this;
+    instance.el.removeAttribute("data-filtering");
+  }
+})();
```
jshawl: this is cool, but it's a lot of code to maintain. Why not replace with another sorting lib? like packery or isotope? both allow custom classes / styling / custom filtering http://isotope.metafizzy.co/

app/assets/javascripts/helpers.js
RobertAKARobin: Just like to keep it light. Could use jQuery, since we're already including it.

app/assets/javascripts/tables.js
RobertAKARobin: DataTables allows custom styling too, but adjusting the HTML itself is more complicated. As we move toward a better UX this will become more important.

RobertAKARobin: Also, I'm pretty sure God's role for me is to rewrite every existing Javascript library.

jshawl: you can do that! recommended action for this: choose another existing pluing

app/assets/javascripts/helpers.js
RobertAKARobin: jQuery doesn't have `collect` and `is_a` methods, though, so would prefer to keep this.

app/assets/javascripts/tables.js
adambray: yeah, my vote for now would be to keep using datatables or a plugin, but @RobertAKARobin I'd like to hear more of your argument and better understand it.

My feelings are related to complexity and maintainability... the library code for sorting is a bit complex, and it doesn't really enable much simplification of the HTML (except, as I can tell, the removal of the `no-sort` class from certain headers).

PR #489, opened 2016-02-22
Moved mantras to separate repo

RobertAKARobin:


RobertAKARobin:
Thought it might be nice to have this in an easier-to-refer-to location for students' benefit. Also, can now update mantras without needing to make PR to Garnet.

jshawl:
if that repo gets renamed, this will 404 and the feature will break. Why not generate the readme from the existing https://garnet.wdidc.org/mantras.json ?

RobertAKARobin:
1. To update the mantras we'd need to create a PR to Garnet and run the barrage of tests.
2. It'd be nice to have the mantras in their own easily-accessible location so we can refer to them as a whole.

nolds9:
:+1: for separation of mantras. Would love for it to be easy for students to submit their own.

adambray:
:+1:

adambray:
I think the extra dependency may be worth the easier contributions of mantras by students.

PR #497, opened 2016-02-23
fixes #418

### LINKED ISSUE #418: "Count Issues" does not show "closed".
After pressing "Count Issues", issue status is always "missing".
http://garnet.wdidc.org/assignments/63?cohort=1

It indicates that Dan's issue is "missing", but it is "closed" (https://github.com/ga-wdi-exercises/pixart_js/issues?q=involves:dan-ator)

jshawl:
should probably replace https://github.com/ga-dc/garnet/blob/ede575b801b5316f0f06a505d5936618741c384e/app/models/assignment.rb#L46 because the output is not human readable

### DIFF
--- a/app/assets/javascripts/assignments.js
+++ b/app/assets/javascripts/assignments.js
@@ -31,8 +31,8 @@ function loadIssues(url, callback){
 
   function parseIssue(issue){
     var userID, state;
-    userID = issue[8][1][1][1];
-    state = issue[10][1];
+    userID = issue[9][1][1][1];
+    state = issue[11][1];
     if(!summaries[userID]) summaries[userID] = {}
     if(!summaries[userID][state]) summaries[userID][state] = 0;
     summaries[userID][state]++


app/assets/javascripts/assignments.js
```diff
@@ -31,8 +31,8 @@ function loadIssues(url, callback){
 
   function parseIssue(issue){
     var userID, state;
-    userID = issue[8][1][1][1];
-    state = issue[10][1];
+    userID = issue[9][1][1][1];
+    state = issue[11][1];
```
RobertAKARobin: Could you add a comment saying what `issue[9][1][1][1]` and `issue[11][1]` return?

jshawl: user id and state?

jshawl: in the variable names!

RobertAKARobin: Doh.

RobertAKARobin:
:+1:

PR #499, opened 2016-02-24
Added autosave

RobertAKARobin:
AJAX method we can use as a catch-all for submitting anything, including submission statuses, scores, notes, attendance, etc...


app/assets/javascripts/application.js
```diff
@@ -60,3 +60,24 @@ new KonamiListener(window, function(){
...
+    url:      url,
+    type:     "post",
+    dataType: "json",
+    data:     data
+  }).fail(function(response){
+    console.log("That didn't work.");
+  }).always(function(response){
+    trigger.removeClass("waiting");
+  });
+}
```
jshawl: :+1:

jshawl:
it would be nice to get some JS test integration with things like this, but let's merge!

PR #501, opened 2016-02-24
Standardize attendances

RobertAKARobin:
Also moves attendance/submission to own partials for extra dryness.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -85,15 +85,6 @@ module ApplicationHelper
     end
   end
 
-  def status_button record, status
-    output = ""
-    id = "#{dom_id(record)}_#{status}"
-    checked = "checked" if record.status == status
-    output += "<input type='radio' name='#{dom_id(record)}' id='#{id}' value='#{status}' #{checked} data-record-url='#{url_for record}' data-record-attribute='status' />"
-    output += "<label for='#{id}' class='status'>#{status}</label>"
-    return output.html_safe
-  end
-
   def cohort_status(cohort, user)
     user.memberships.find_by(cohort: cohort).status.to_s
   end
--- a/app/views/components/_status_buttons.html.erb
+++ b/app/views/components/_status_buttons.html.erb
@@ -1,0 +1,10 @@
+<% klazz = record.model_name.to_s.classify.constantize %>
+<%= form_for record, html:{id: nil}, data:{autosave: true} do |f| %>
+  <div class="buttons">
+    <% klazz.statuses.keys.each do |status| %>
+      <% id = "#{record.id}-status-#{status}" %>
+      <%= f.radio_button :status, status, id: id %>
+      <%= f.label :status, status, for: id %>
+    <% end %>
+  </div>
+<% end %>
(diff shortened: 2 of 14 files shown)


app/helpers/application_helper.rb
```diff
@@ -85,15 +85,6 @@ def markdown(text)
...
-  def status_button record, status
-    output = ""
-    id = "#{dom_id(record)}_#{status}"
-    checked = "checked" if record.status == status
-    output += "<input type='radio' name='#{dom_id(record)}' id='#{id}' value='#{status}' #{checked} data-record-url='#{url_for record}' data-record-attribute='status' />"
-    output += "<label for='#{id}' class='status'>#{status}</label>"
-    return output.html_safe
-  end
-
   def cohort_status(cohort, user)
```
andrewsunglaekim: I've always been confused whether to use helpers vs partials. Is there any kind of rule or convention?

RobertAKARobin: Yeah, I dunno. After looking back at this, it seems to me that if something involves writing a bunch of HTML then it's better as a partial.

app/views/components/_todo_table.html.erb
```diff
@@ -0,0 +1,9 @@
+<% class_name   = collection.model_name %>
+<%= render partial: "#{class_name.plural}/thead" %>
+<% collection.each do |record| %>
+  <%= render partial: "#{class_name.plural}/tr", locals: {
+    :"#{class_name.singular}"   => record,
```
andrewsunglaekim: whoa, this is really cool, i had no idea you could make dynamic partials like this... this is sick.

app/views/submissions/_tr.html.erb
```diff
@@ -18,32 +21,22 @@
...
-            <% id = "#{submission.id}-status-#{status}" %>
-            <%= f.radio_button :status, status, id: id %>
-            <%= f.label :status, status, for: id %>
-          <% end %>
-        </div>
+        <%= f.text_field :score, class: "sm", id: nil %><%= out_of_score %>
       <% end %>
     <% else %>
-      <%= submission.status %>
+      <%= "#{submission.score}#{out_of_score}" %>
```
andrewsunglaekim: this feels like it should be a model method for submissions

RobertAKARobin: Could be... I only envision it showing up in HTML, though.

app/views/components/_status_buttons.html.erb
```diff
@@ -0,0 +1,10 @@
+<% klazz = record.model_name.to_s.classify.constantize %>
+<%= form_for record, html:{id: nil}, data:{autosave: true} do |f| %>
```
andrewsunglaekim: el o el

andrewsunglaekim:
this is a really cool PR, i don't know enough about conventions for these things for my thumbs up to hold much value but :+1: from me. My comments are more just cause i'm curious

app/helpers/application_helper.rb
andrewsunglaekim: yeah, taht sounds right

RobertAKARobin:
Thanks, man! I really appreciate it!

amaseda:
:+1:

Issue #495, opened 2016-02-23
Add number of observations and date of most recent observation to cohort show page


RobertAKARobin:

andrewsunglaekim:
on a per student basis in the members section?

PR #505, opened 2016-02-26
adds observation feed to instructor dashboard

andrewsunglaekim:
Couple of things that I was unsure about, but here's the first PR. Would like feedback on:

### Partials changed
Changed the partial to include the users name of that membership. Only problem is now for the observations on the membership's show page will have the same user name. 

Potential Solutions:
- some if/else conditionals that obfuscate users name for membership show page
- different partial for observation feed

### UI and how it looks
Please feel free to change this, no idea what it should look like by the end, maybe a sidebar?

### Turned off caching for development
Maybe this should go in another PR or need to be a discussion.

### Spec
The test is marginal at best, would like to include another cohort that currently isn't in session but has memberships with observations to round it out. Also updated the dashboard setup to include variables to point at memberships


RobertAKARobin:
Instead of being on the dashboard page, which is already a little crowded, could you add a "most recent observation date" column to the cohort show page? It would be automatically searchable/filterable thanks to the dataTables thing.

https://github.com/ga-dc/garnet/issues/495

RobertAKARobin:
Yeah, I think turning off caching should be its own PR

RobertAKARobin:
#scrutiny

andrewsunglaekim:
i didn't wanna mess with the cohort show page for that reason.. lol

RobertAKARobin:
The filtering/searching is just front-end Javascript, so it shouldn't hurt anything.

My biggest worry is just slowing down/cuttering up the dashboard. We already load all the observation data on the cohort show page anyway, so may as well use it...?

adambray:
Yeah, I vote to moving this to the cohort show page or somewhere else. My main reasons are:

1. Keeping the dashboard clearly scoped to "things I need to do"
2. Observations are sensitive, and shouldn't be visible on your screen without explicitly looking at them, in case you go to garnet home page while on a projector, sitting next to a student, etc.

jshawl:
@andrewsunglaekim want to move to cohort show?

andrewsunglaekim:
yeah, i can do that

andrewsunglaekim:
Most recent changes reverts caching to be enabled. Moves feed into separate view. Adds link to cohort show page. Adds new spec.

app/views/observations/_observation.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="observation <%= color_of_status(observation.status) %>" id="observations-<%= observation.id %>">
   <%= markdown(observation.body) %>
-  <p><small>By
+  <p><small><%= link_to observation.membership.user.name, observation.membership %> By
```
adambray: This can just be `observation.membership.name`, since name is delegated from membership to user.

app/controllers/cohorts_controller.rb
```diff
@@ -90,6 +90,9 @@ def generate_invite_code
     redirect_to :back
   end
 
+  def observations
+    @observations = Observation.joins(:membership).where("memberships.cohort_id = ?", @cohort.id).order(created_at: :desc).limit(10)
```
adambray: Wondering if a cohort should have many observations through memberships...

andrewsunglaekim: Can't think of a use for that scope other than this feature, worth?

adambray: Not technically a scope (really an association)... but honestly I'm on the fence as well. Just wanted to start a discussion.

adambray:
:+1: pending replacing membership.user.name with membership.name.

andrewsunglaekim:
changed to use delegate. Also replaced the delegate method with a delegate helper, is this ok? eliminated 2 lines of code, specs still pass using it

spec/features/observation_feed_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  scenario 'should see recent observations from all admins in cohort' do
+    login_user(test_instructor)
+    visit "/cohorts/#{cohort1.id}/observations"
+    expect(page).to have_content "johns obs from test instructor"
+    expect(page).to have_content "janes obs from other test instructor"
+  end
+
+  scenario 'should not see observations from other cohorts' do
```
adambray: :exclamation: :exclamation: :exclamation: :exclamation: 
this test is broken @andrewsunglaekim . The controller code does no authorization checked, and allows anyone to see observations for any cohort.
:exclamation: :exclamation: :exclamation: :exclamation: 

cc @ga-dc-bug-bounty

spec/features/observation_feed_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  scenario 'should see feed header' do
+    login_user(test_instructor)
+    visit "/cohorts/#{cohort1.id}/observations"
+    expect(page).to have_content "Last 10 Observations"
+  end
+
+  scenario 'should see recent observations from all admins in cohort' do
+    login_user(test_instructor)
+    visit "/cohorts/#{cohort1.id}/observations"
```
adambray: Not a bug, but an improvement since you'll need to update this file anyway...

``` 
visit "/cohorts/#{cohort1.id}/observations"
# better as
visit observations_cohort_path(cohort1)
```

spec/features/observation_feed_spec.rb
adambray: (By broken, I mean it's not testing what it claims to be testing, and as a result the bug in the controller was not detected)

adambray: I'm working on a fix right now.

andrewsunglaekim: oooh yeah, thats not good

spec/features/observation_feed_spec.rb
```diff
@@ -12,6 +12,7 @@
 
   let(:john) { User.find_by(username: 'john') }
   let(:jane) { User.find_by(username: 'jane') }
+  let(:dave) { User.find_by(username: 'jane') }
```
adambray: Crap this is a bug.

adambray:
@RobertAKARobin @mattscilipoti @jshawl @andrewsunglaekim 

I think this is ready for a final review now.

nolds9:
Ok on the concept, will review the implementation

RobertAKARobin:
Aww snap, it finally happened!

PR #510, opened 2016-03-02
Standardized headers

RobertAKARobin:


nolds9:
:+1:

adambray:
Overall I think this is an improvement, so :+1: happy if this gets merged in as it is now. Some thoughts:

I'd vote to keep the header for Cohorts (e.g. the Cohort name) when on the cohort show page, and the same thing for events and assignments. It feels not quite header-y enough in the new style.

I guess what I'm really trying to say is I think this feels close to breadcrumb nav, but not quite, in a way that makes it feel a bit awkward to me. What if we tried real breadcrumb navigation?

`DC > WDI 8 > Assignments > Pixart`

`DC > WDI 9 > Johnny Drop Tables`

adambray:
Oops, didn't mean to close it!

RobertAKARobin:
I'm into that! Give me a minute

RobertAKARobin:
@adambray I agree! I'd like to merge in https://github.com/ga-dc/garnet/pull/514 before that.

PR #511, opened 2016-03-02
Removed folding in various places

RobertAKARobin:
The pages really aren't that long without it. I think it's more of a nuisance having to expand everything all the time.


nolds9:
Personally, I like the folding. But ok either way

amaseda:
:+1: Cool with this.

PR #513, opened 2016-03-02
Fixes 508 by touching memberships after user update

adambray:


app/models/user.rb
```diff
@@ -134,4 +134,8 @@ def accept_invite
     end
   end
 
+  def touch_memberships
+    memberships.each(&:touch)
```
nolds9: had to look into what `touch` does. Seems like a good use-case, but maybe include a code comment since you're using short hand syntax...

nolds9:
:+1:

andrewsunglaekim:
yeah, this looks good but CI is failing

adambray:
I'll take a look at this

adambray:
Opened new PR with passing tests:
https://github.com/ga-dc/garnet/pull/607 

Closing this one.

PR #514, opened 2016-03-02
Added cohort number at location

RobertAKARobin:
Ex: If you have DC WDI 8, makes 8 an actual column.


adambray:
@RobertAKARobin no issues with the code here, but what's your motivation for this change... might be good to check with some producers if this is helpful... reason I say that is I've heard that some campuses don't track courses numerically. But maybe that's just a student-facing change, and we are all still tracking them numerically internally.

RobertAKARobin:
@adambray I think this is an optional field that campuses can use as they want. My motivation was breadcrumbs. I'd like to have something like we used to:

```
DC > WDI > 8
```

As opposed to:

```
DC > WDI > DC_WDI_8
```

...or manipulating the string each time. If a course doesn't have a number, then the full name can be used instead. It may also be helpful for campus' tracking purposes.

adambray:
:+1:

Issue #519, opened 2016-03-06
assignments unordered on cohort show page


jshawl:
should be ordered by creation date and/or filterable by due date or name

Issue #515, opened 2016-03-03
Set up jsDataTables for additional tables


RobertAKARobin:
Particularly:
- `events/_table.html.erb`
- `assignments/_table.html.erb`

I've tried adding it to these and get an error every time.

Also, it would be nice to have it compatible with the Memberships table, which has `rowspan` in it.

PR #521, opened 2016-03-07
Added custom table sorting JS library

RobertAKARobin:
- Sorting
  - Not limited to tables: can be used for any series of records
  - Also includes filtering, both by typing input and clicking "trigger" elements (e.g. tags)
  - Angular-ish HTML attribute-based system
  - Updated all tables to match
  - Works with `colspan` and `rowspan`, unlike `jsDataTables`
- Also moved tag_list to its own partial view

- Fixes https://github.com/ga-dc/garnet/issues/519
- Fixes https://github.com/ga-dc/garnet/issues/515


adambray:
@RobertAKARobin overall, I really like this! I'm giving it my :+1: pending a bit more testing I'd like to do tomorrow morning.

Only one thought... how hard would it be to make it so we don't have to specify `data-sort-trigger="foo"` and `data-sort-field="foo"`? I.e. to just place a `data-sort-trigger` on the header, and if no value is specified, to just sort by that column (while still supporting the current behavior of letting you specify the field if need be.

RobertAKARobin:
I think it'd require writing a lot more code: you'd need to count the index of that particular header, find the matching cell, and *then* start to consider `colspan`. That's the approach I tried a couple times before settling on this.

RobertAKARobin:
Thanks for reviewing!

adambray:
:+1: 

I'm ok with merging this in as long as we create an issue for the multi-column sort (it can be a low priority one).

jshawl:
:+1: pending issue creation and merge conflict resolution. good to merge!

PR #527, opened 2016-03-08
fixes #524

### LINKED ISSUE #524: % present missing from users show page

jshawl:


RobertAKARobin:
Thumb sideways. This means we can't sort by tardy anymore.

adambray:
I'm not a fan of this. To me knowing the spread of tardies vs absences is far more important.

adambray:
Also, these numbers are necessary to determine graduation status (given the formula for tardies -> absences).

jshawl:
Student is requesting % present. Should I tell them no on that feature?

RobertAKARobin:
Maybe let's just have % present instead of % absent

adambray:
porque no los tres? (mostly jk)

Why do they want it? That's the important question to me.

-Adam

RobertAKARobin:
See https://github.com/ga-dc/garnet/pull/530

PR #532, opened 2016-03-09
show todos for all instructors

jshawl:


app/views/cohorts/todos.html.erb
```diff
@@ -0,0 +1,26 @@
+<h2>Number Assignments ungraded</h2>
+<div class='chart todos-chart'></div>
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.16/d3.min.js"></script>
```
adambray: should we vendor this?

app/views/cohorts/todos.html.erb
```diff
@@ -0,0 +1,26 @@
...
+<script>
```
adambray: Can this be moved into a js file?

app/assets/javascripts/todos.js.erb
```diff
@@ -0,0 +1,22 @@
...
+    .data(todoData)
+    .enter()
+    .append("div")
+    .style("width", function(d){
+      return scale(d.ungraded) + "%"
+    })
+    .text(function(d){
+      return d.user + ": " + d.ungraded
+    })
+}
```
RobertAKARobin: Does this need to be an ERB?

RobertAKARobin:
:+1: pending tests!

PR #534, opened 2016-03-09
Changed Observation status integers

RobertAKARobin:
Now they better match Submissions and Attendances, which have 0 as "unmarked". For Observations, "neutral" changes from 3 to 0.

### DIFF
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -89,8 +89,9 @@ class Membership < ActiveRecord::Base
   end
 
   def update_average_observations
-    average = self.observations.where.not(status:3).average(:status).to_f.round(2)
-    self.update!(average_observations: self.observations.any? ? average : nil)
+    obz = self.observations.where.not(status: Observation.statuses[:neutral])
+    average = obz.any? ? obz.average(:status).to_f.round(2) : 0
+    self.update!(average_observations: average)
     return average
   end
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160318191715) do
+ActiveRecord::Schema.define(version: 20160322171539) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -77,6 +77,14 @@ ActiveRecord::Schema.define(version: 20160318191715) do
     t.integer  "cohort_id"
   end
 
+  create_table "groups", force: :cascade do |t|
+    t.string   "title"
+    t.string   "category"
+    t.integer  "parent_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
   create_table "locations", force: :cascade do |t|
     t.string   "name"
     t.string   "short_name"
--- a/spec/models/membership_spec.rb
+++ b/spec/models/membership_spec.rb
@@ -5,9 +5,6 @@ RSpec.describe Membership do
     load "#{Rails.root}/db/seeds/test_seed.rb"
     cohort = Cohort.last
     @m = cohort.memberships.create()
-    @m.observations.create(status: 0)
-    @m.observations.create(status: 0)
-    @m.observations.create(status: 1)
     cohort.events.create!(title:"Some test event2", occurs_at: Time.now)
     @m.attendances.first.present!
     cohort.events.create!(title:"Some test event", occurs_at: Time.now + 3.hours)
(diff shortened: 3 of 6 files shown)


db/schema.rb
```diff
@@ -73,6 +73,14 @@
     t.integer  "cohort_id"
   end
 
+  create_table "groups", force: :cascade do |t|
```
RobertAKARobin: wtf

RobertAKARobin:
Fixes https://github.com/ga-dc/garnet/issues/538

adambray:
@RobertAKARobin I want to give this some more thorough testing, but I wanted to point out a quick issue:

See here and note the magic number '3' being used in the code:
https://github.com/ga-dc/garnet/blob/master/app/models/membership.rb#L92

Can you write a test that makes the current version of this PR fail? And then fix the code to make it pass by not using a magic number? If you don't have the bandwidth, I can do so tonight.

adambray:
pinging @jshawl @andrewsunglaekim @tylercrosse so they can learn from this (it's a common class of mistake in programs).

http://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad

adambray:
Also, this means the scaling factor will need to be changed in `Membership#average_observations`. As of this PR, it's returning values from 1.0 - 3.0, rather than 0.0 - 2.0 as before.

adambray:
I'm totally in favor of this PR's goal though!

spec/models/membership_spec.rb
```diff
@@ -47,7 +47,7 @@
     end
     it "has an average observation not influenced by neutral status" do
       @m.observations.create(status: 3)
```
adambray: Another leftover magic number from the previous PR, should be changed to `Observation.statuses[:good]`.

spec/models/membership_spec.rb
```diff
@@ -47,7 +47,7 @@
...
-      expect(@m.average_observations).to eq(0.33)
+      expect(@m.average_observations).to eq(2)
```
adambray: How is this test passing? If the membership has one with status `3`, how is the average returning `2`? 

(I don't see where the subtraction by one is happening)

RobertAKARobin: Here: https://github.com/ga-dc/garnet/blob/standardize-obervation-statuses/spec/models/membership_spec.rb#L10

spec/models/membership_spec.rb
RobertAKARobin: Fixed: https://github.com/ga-dc/garnet/commit/ebd1e28300dc856230b3446b9aec5b426d172a6f#diff-d537d310888ce2fead5da6d283c35eaeR49

RobertAKARobin: @adambray Fixed: https://github.com/ga-dc/garnet/commit/ebd1e28300dc856230b3446b9aec5b426d172a6f#diff-d537d310888ce2fead5da6d283c35eaeR49

adambray: The fact that the test was passing, even though it appears to be incorrect was a smell to me, and I think the test was broken. This was due to the fact that there were other magic numbers in the code (particularly the test setup for this set of tests). I'm pushing a commit that demonstrates this. 

Note that I replaced the statue numbers with what they would represent pre-change.

adambray:
Note the tests now fail as expected. This is because the average_observations method needs to be updated accordingly to subtract 1.

RobertAKARobin:
Ah, sorry. I should have fixed the other magic numbers

adambray:
No worries. How I see myself when code reviewing:

hacker

app/models/membership.rb
```diff
@@ -89,8 +89,9 @@ def average_observations
   end
 
   def update_average_observations
-    average = self.observations.where.not(status:3).average(:status).to_f.round(2)
-    self.update!(average_observations: self.observations.any? ? average : nil)
+    obz = self.observations.where.not(status: Observation.statuses[:neutral])
```
andrewsunglaekim: that's my b on the magic number

RobertAKARobin:
Tested with production data.

adambray:
:+1:

db/migrate/20160309162046_make_neutral_observations_zero.rb
```diff
@@ -0,0 +1,14 @@
+class MakeNeutralObservationsZero < ActiveRecord::Migration
+  def change
+    Observation.all.each do |observation|
+      next if !observation.membership || !observation.user
+      status_in = observation.read_attribute("status")
+      if status_in === 3
```
andrewsunglaekim: this is nil when Jesse ran the migration

adambray: can you clarify what you mean here?

andrewsunglaekim: that prevents the migration from running. Maybe best to add:

```
if status_in === nil
status_in = 0
```

andrewsunglaekim:
@RobertAKARobin see comments about the migration and possible solutions. Other than that pair reviewed with Jesse and looks good pending the fix for nil statuses on observations.

RobertAKARobin:
@andrewsunglaekim This should be good now.

tylercrosse:
Related to #538

db/migrate/20160316171539_make_neutral_observations_zero.rb
```diff
@@ -0,0 +1,14 @@
+class MakeNeutralObservationsZero < ActiveRecord::Migration
+  def change
+    Observation.all.each do |observation|
+      next if !observation.membership || !observation.user
+      status_in = observation.read_attribute("status")
+      if status_in === 3 || status_in.nil?
```
andrewsunglaekim: :+1:

andrewsunglaekim:
@RobertAKARobin  last thing i promise!!!!! anyway we can get a constraint or validation to not allow for nil observations?

RobertAKARobin:
This line checks whether it's equal to 3 or nil:

https://github.com/ga-dc/garnet/pull/534#discussion-diff-56426623R6

andrewsunglaekim:
i think this ones good to merge, paired with jesse and the nil thing got fixed so :+1:

PR #539, opened 2016-03-09
Refactored coloring of statuses and percents (Contingent on standardizing Obz statuses)

RobertAKARobin:
- A single "#color_of" helper method that takes a variety of inputs
- Colors are rendered with `style` tags on HSL scale instead of with CSS classes
- Specs included
- We won't talk about how much math this took the first time around

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -1,0 +1,93 @@
+require "rails_helper"
+
+RSpec.describe ApplicationHelper do
+  describe "#color_of" do
+    # These are defined using the HSL color spectrum
+    let(:red){ 0 }
+    let(:yellow){ 60 }
+    let(:green){ 120 }
+    let(:orange){ 48 }
+    context "when passed a submission or attendance" do
+      before(:all) do
+        u = User.create(name: "test")
+        c = Cohort.create
+        @sub = c.add_member(u).submissions.create
+        @options = {type: "int"}
+      end
+      it "returns empty string for unmarked" do
+        @sub.update!(status: "unmarked")
+        expect(color_of(@sub, @options)).to eq("")
+      end
+      it "returns red for missing" do
+        @sub.update!(status: "missing")
+        expect(color_of(@sub, @options)).to eq(red)
+      end
+      it "returns yellow for incomplete" do
+        @sub.update!(status: "incomplete")
+        expect(color_of(@sub, @options)).to eq(yellow)
+      end
+      it "returns green for complete" do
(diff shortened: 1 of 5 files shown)


spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+        @sub.update!(status: "missing")
+        expect(color_of(@sub, @options)).to eq(0)
+      end
+      it "returns yellow for incomplete" do
+        @sub.update!(status: "incomplete")
+        expect(color_of(@sub, @options)).to eq(60)
+      end
+      it "returns green for complete" do
+        @sub.update!(status: "complete")
+        expect(color_of(@sub, @options)).to eq(120)
```
andrewsunglaekim: where are these numbers coming from the equals expectations of `0, 60, 120`

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      end
+      it "returns red for 80%" do
+        expect(color_of(80, @options)).to eq(0)
+      end
+      it "returns red for 70%" do
+        expect(color_of(70, @options)).to eq(0)
+      end
+    end
+  end
+end
```
andrewsunglaekim: is there a way to test for an irrational average obseration?

spec/helpers/application_helper_spec.rb
RobertAKARobin: They're HSL hue colors: `0` is red, `120` is green.

spec/helpers/application_helper_spec.rb
RobertAKARobin: You sure you mean irrational numbers? That's like `√-i`

andrewsunglaekim: i guess what im asking is say we have 3 observations 2 green 1 red that will return something like 4/3 how does that translate into a hue and can we test for that?

RobertAKARobin:
@andrewsunglaekim Tests added! https://github.com/ga-dc/garnet/pull/539/files#diff-31c50b16a4c9f23be71c49d30afcf7edR75

andrewsunglaekim:
yeah, this looks great. I want to merge it in, but honestly, when i looked at it myself, and then with jesse, i still wasn't exactly sure what was going on. I have checked out the branch and see that it works though! :+1: from me but would love @adambray || @mattscilipoti  to pair on this #tooadvanced...yet?

jshawl:
:+1:

PR #545, opened 2016-03-10
Add endpoint for submission from outcomes

jshawl:


jshawl:
chrome extension if you wanna peek https://github.com/ga-dc/outcomes-garnet-bridge/blob/master/src/inject/inject.js

Gemfile
```diff
@@ -4,6 +4,7 @@ source 'https://rubygems.org'
 gem 'rails', '4.2.4'
 # Use postgresql as the database for Active Record
 gem 'pg'
+gem 'thin'
```
adambray: what's the purpose of this change in the context of this PR?

app/controllers/api/observations_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    def from_outcomes
+      membership = Membership.find_by(outcomes_id: params[:id])
+      body = membership.observations.create(body: params[:body], admin_id: current_user.id)
+      render json: body
+    end
+
+    private
+    def allow_cors
+      headers['Access-Control-Allow-Origin'] = '*'
```
adambray: Can we test limiting this to localhost, *.wdidc.org, and *.generalassemb.ly?

Gemfile
jshawl: allows us to run https server locally `thin start --ssl`

app/controllers/api/observations_controller.rb
jshawl: I'm able to curl localhost even with `headers['Access-Control-Allow-Origin'] = 'http://google.com'
`

do you have example code anywhere?

config/routes.rb
```diff
@@ -61,9 +61,11 @@
   end
   resources :observations,  only: [:create, :destroy]
 
+
   namespace :api do
     resource :user, only: [:show]
     get 'send_api_token'
+    match 'observations/from_outcomes' => 'observations/from_outcomes', :via => :post
```
RobertAKARobin: Didn't know this existed

app/controllers/api/observations_controller.rb
jshawl: any ideas @RobertAKARobin ?

RobertAKARobin: Afraid not. I don't know much about CORS, especially on Rails.

jshawl: screenshot 2016-03-16 10

app/controllers/api/observations_controller.rb
```diff
@@ -0,0 +1,20 @@
+module Api
+  class ObservationsController < ApplicationController
+    before_filter :allow_cors, only: [:from_outcomes]
+    skip_before_action :authenticate
```
adambray: Can we remove this?

app/controllers/api/observations_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    def from_outcomes
+      membership = Membership.find_by(outcomes_id: params[:id])
```
adambray: This param should be called `params[:outcome_id]` for clarity.

adambray: we should have authorization code like `if can manage membership`

app/views/memberships/show.html.erb
```diff
@@ -52,3 +52,11 @@
     <%= link_to "Refresh Github info", gh_refresh_user_path(@user), class: "button" %>
   <% end %>
 </section>
+
+<% if can? :manage, @cohort %>
```
adambray: This would read better (and have the same effect) if we used `can? :manage, @admin` right?

config/routes.rb
adambray: not sure, but can't this be shortened to:

```
post 'observations/from_outcomes'
```

adambray: Yeah, just checked and `match` is best used when you need to route multiple verbs with the same path to the same controller / action.

adambray:
:+1: :+1:  from Adam and Joe.

PR #546, opened 2016-03-11
Updates dependancies, moves webconsole to dev only

adambray:


adambray:
I moved web console because the latest version throws an error telling you not to load it in a test environment (if it detects that is is).

andrewsunglaekim:
:+1: paired with jesse

RobertAKARobin:
:+1: This means we can merge, right?

jshawl:
yep!

PR #549, opened 2016-03-11
Admin ability to make assignment public/private to students, including tests

tylercrosse:
Assignments are private by default. This PR adds a feature to make them public to students by checking a checkbox on assignment creation / edit. Including tests

Addresses issue #522


jshawl:
:+1: :+1: paired with adrian

db/migrate/20160311152448_add_another_column_to_assignment.rb
```diff
@@ -0,0 +1,5 @@
+class AddAnotherColumnToAssignment < ActiveRecord::Migration
+  def change
+    add_column :assignments, :public, :boolean
```
adambray: can we set a default value on this? I suspect `true` is best from a UX standpoint.

adambray:
:+1: pending addition of default value for public boolean.

andrewsunglaekim:
@tylercrosse could you implement default value of the boolean, then we can merge this bad boy in!

tylercrosse:
Based on the discussion of using this as an internal tool think it makes more sense for grades to be private by default. By forcing admins to specify that an assignment score be public we avoid accidentally revealing info that we don't mean to.

andrewsunglaekim:
yeah i think thats fine. go ahead an implement default value to be private

tylercrosse:
Cool yeah, I made it so it's `:public, default => false`  a.k.a. private :dizzy_face: It should be good to Merge the PR

Issue #554, opened 2016-03-15
Overriding GitHub Profile Info


timfoley:
I'd like to be able to edit student names and override what is auto-populated by GitHub. Some of my students haven't put their full names into their profiles so they just show up as `clewinth` or `ragarza`. I've had students concerned about privacy and reluctant to put their names onto online profiles in the past, and in general, it'd just be easy for me to update their names myself.

jshawl:
pseudocode:

```
current_user.cohorts.each do c
  c.memberships.each do m
    if current_user admins m
      current_user can edit user profile
    end
  end
end
```

PR #555, opened 2016-03-15
allow editing other users; re #554

jshawl:


jshawl:
:+1: becky and joe wrote code and tests

Gemfile
```diff
@@ -23,6 +23,7 @@ gem 'octokit'
 gem 'redcarpet' # use redcarpet for markdown support
 gem 'sass-rails', '~> 5.0' # Use SCSS for stylesheets
 gem 'select2-rails'
+gem 'thin'
```
jshawl: should be moved to development

nolds9:
:+1:

PR #563, opened 2016-03-19
Metrics: Sandi Metz rules via sandi_meter, outputs to /metrics/sandi_meter

mattscilipoti:
Using html output from sand_meter:  https://github.com/makaroni4/sandi_meter

- `rake -T metrics`
- cron job updates nightly, see config/schedule.rb
- sandi_meter outputs to "/metrics/sandi_meter"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@
 /config/application.yml
 /public/assets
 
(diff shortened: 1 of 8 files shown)


nolds9:
:+1:

PR #571, opened 2016-03-24
metrics: add nav link and fix asset paths

mattscilipoti:


RobertAKARobin:
Looks good to me!

lib/tasks/metrics.rake
```diff
@@ -24,6 +24,15 @@ namespace :metrics do
       logger.info message
       # puts message
       `sandi_meter --graph --quiet --output-path "#{sandi_pages_path}"`
+
+      # update assets for rails public dir
+      index_file = sandi_pages_path.join('index.html')
+      text = File.read(index_file)
+      # add leading slash
+      new_contents = text.gsub(/href="assets/, 'href="/assets')
+      new_contents = new_contents.gsub(/src="assets/, 'src="/assets')
```
jshawl: does this work? fixed if you have trailing slash https://garnet.wdidc.org/metrics/sandi_meter/

mattscilipoti: That's interesting.
Since we are referencing the sand_meter page from within the public dir, it wasn't using the asset pipeline, therefore  `src="assets/jquery.js"` needed to be `src="/assets/jquery.js"`.  I guess the trailing slash, in the url, acts as a leading slash for the asset reference.
I tested both urls (w/ and w/o trailing slash) locally, with this fix.

jshawl:
:+1:

PR #574, opened 2016-03-27
Iss383 sidebar

tylercrosse:
Currently lacks tests, but getting closer to being able to merge. I would love some feedback :eyes:

### DIFF
--- a/app/views/cohorts/show.html.erb
+++ b/app/views/cohorts/show.html.erb
@@ -1,3 +1,13 @@
+<nav class="page-nav navbar-relative">
+  <ul>
+    <li><a href="#top">Top</a></li>
+    <li><a href="#members">Members</a></li>
+    <li><a href="#attendance">Attendance</a></li>
+    <li><a href="#assignments">Assignments</a></li>
+  </ul>
+</nav>
+<div class="nav-spacer">&nbsp;</div>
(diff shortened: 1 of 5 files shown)


app/views/cohorts/show.html.erb
```diff
@@ -1,3 +1,13 @@
+<nav class="page-nav navbar-relative">
```
nolds9: I think a style guide we are trying to maintain is that any classes that are used for javascript selection should be prefixed with `js`, i.e. `js-page-nav`

nolds9:
:+1: Really cool Tyler! Some initial ux thoughts: it would be great if we could consolidate the navigation so there's not 3 separate nav bar's. Also possibly we be cool to have a way to "toggle" the navigation, that is add an icon and functionality to expand and collapse it. But these are all nice to haves, would be happy to have it in as is.

tylercrosse:
I think this is good enough to be useful, would love it to be looked over and merged in.

app/views/cohorts/show.html.erb
andrewsunglaekim: what if its for both CSS && JS? Might be ok

andrewsunglaekim:
:+1:

PR #579, opened 2016-04-06
fixes #543, updates percent_from status

### LINKED ISSUE #543: Why is hw_complete all 0?
Maybe solution is a "refresh all averages" button.

andrewsunglaekim:


app/models/assignment.rb
```diff
@@ -21,8 +21,8 @@ class Assignment < ActiveRecord::Base
   ]
 
   def create_submissions
-    self.cohort.memberships.where(is_admin: false).each do |membership|
-      membership.submissions.unmarked.create(assignment_id: self.id)
+    self.cohort.memberships.student.active.each do |membership|
```
adambray: This line doesn't make the fix, but it's cleaner, and prevents creating submissions for students who are deactivated.

app/models/assignment.rb
```diff
@@ -21,8 +21,8 @@ class Assignment < ActiveRecord::Base
...
+      membership.submissions.create(assignment_id: self.id, status: Submission.statuses[:unmarked])
```
adambray: This is the actual fix. The error is that the previous line was locking in the `unmarked` scope on a given membership's `submissions`  method. This would cause the `update_percents` method to think there were no marked submissions and thus cache `0` as the values.

app/models/membership.rb
```diff
@@ -60,11 +60,11 @@ def update_percents_of klass
     # = submissions
     related_records   = self.send(klass_plural_name).due
     # = self.submissions.due
-    unmarked_records  = related_records.where.not(status: klass.statuses[:unmarked])
+    marked_records  = related_records.where.not(status: klass.statuses[:unmarked])
```
adambray: This is just changing the variable name to be semantically correct (before it was the opposite of what it held)

app/models/assignment.rb
mattscilipoti: Recommend a comment here, indicating why we aren't using the scope.  I can totally see me "correcting" this oversight in a future commit.  Maybe a spec to help catch us too?  I suggest a spec at the place where we identified the issue (where we needed ALL of this member's submissions), rather than for this changed code.  I'd be happy help work through that.

PR #583, opened 2016-04-19
Brakeman: security checker

mattscilipoti:
Adds brakeman to /metrics.
- outputs to "public/metrics/brakeman"
- https://github.com/presidentbeef/brakeman

To test manually, use rake tasks: `rake -T metrics`

Note: immediately after pushing, you will receive "No route matches [GET] "/metrics/brakeman" because the report as not been generated (public/metrics/brakeman/index.html).  You can either generate the report, via the rake task, or wait for the nightly cron/rake task.

### DIFF
--- a/app/models/metric.rb
+++ b/app/models/metric.rb
@@ -1,0 +1,7 @@
+require 'logger'
+
+class Metric
+  def self.names
+    %w(brakeman sandi_meter).sort
+  end
+end
--- a/app/models/metric.rb
+++ b/app/models/metric.rb
@@ -1,5 +1,3 @@
-require 'logger'
-
 class Metric
   def self.names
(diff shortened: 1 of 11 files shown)


app/models/metric.rb
```diff
@@ -0,0 +1,7 @@
+require 'logger'
```
adambray: Is this required?

adambray:
👍

app/models/metric.rb
mattscilipoti: Not any more.  Removing.

nolds9:
needs team review

RobertAKARobin:
We don't know what this does. Can re-open later if we find a need.

PR #585, opened 2016-04-23
cohort: convert methods to associations

mattscilipoti:
- converts Array results to AR Associations
- user factory derives email from username

This is preparation for dealing with Instructors in a Cohort.


adambray:
👍 Looks good to me!

PR #586, opened 2016-04-23
members#show: fold all headers

mattscilipoti:
- All headers on Member show page default to folded.
- Allows you to fold and unfold the header


mattscilipoti:
Closing.  I want to create a section tag helper and reuse that.

PR #591, opened 2016-04-27
Fixes # 551

andrewsunglaekim:
Couple of notes:

- used a setter instead of `.unmarked!` because the enum bang method fires a save
- test uses a hard coded string, while i didn't like this, couldn't figure out an easy way to get the string of the key for a certain value in the `Attendance.statuses` hash. It ended up being much cleaner and more englishy hard coding the string.


jshawl:
👍

PR #592, opened 2016-04-29
Active loading

nolds9:
Used `bullet` gem to find extra `n+1` queries, fixed by utilizing eager-loading

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -38,4 +38,27 @@ Rails.application.configure do
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+
+  config.after_initialize do
+    Bullet.enable = true
+    Bullet.alert = true
+    Bullet.bullet_logger = true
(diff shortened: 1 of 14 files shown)


config/environments/development.rb
```diff
@@ -38,4 +38,27 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+
+  config.after_initialize do
+    Bullet.enable = true
+    Bullet.alert = true
```
nolds9: Can deactivate in development if alert becomes annoying

jshawl:
what does this mean?!

jshawl:
Is there a test we can write to show this optimization? if not that's cool 👍

nolds9:
Doubt there's an easy way to test it considering its just measuring queries. Here's a little demo if interested: https://github.com/flyerhzm/bullet#demo

nolds9:
@adambray

PR #595, opened 2016-04-29
Issue#593 Updates observation_average when observation is deleted

awhit012:


awhit012:
When I went back and wrote a test, it forced me to rethink it and I rewrote the solution. Another reason why testing is great. :)

nolds9:
👍  & tests!

awhit012:
Yeah I submitted earlier without the test, and Adam asked if I would write a test for it. Went back to write the test and that made me figure out a better way to solve the issue! That made me pretty happy.

amaseda:
:+1:

PR #607, opened 2016-05-16
Fixes #508 - updates memberships when users are updated - take 2

### LINKED ISSUE #508: Membership index does not update User first/last names after pulling from github.
Maureen V (maureenv) updated her github info this morning.  After "Refreshing all github info", the Membership listing is still using her username (https://garnet.wdidc.org/cohorts/4#members).  Her show page has her first/last name.

adambray:
An updated version of [this PR](https://github.com/ga-dc/garnet/pull/513) to make tests pass.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -137,5 +137,5 @@ class User < ActiveRecord::Base
   def touch_memberships
     memberships.each(&:touch)
   end
-  
+
 end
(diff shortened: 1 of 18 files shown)


app/models/user.rb
```diff
@@ -134,4 +134,8 @@ def accept_invite
     end
   end
 
+  def touch_memberships
+    memberships.each(&:touch)
```
nolds9: why do we have to update all of a user's memberships when a user itself is updated?

adambray: It's for caching purposes. The caching is all done on a membership level, but if a user's name is update, the cached membership row views should be updated as well

PR #530, opened 2016-03-08
Tardies shown as number rather than percent

RobertAKARobin:
- TODO: Add coloring
- Changes % absent to % present
- Fixes https://github.com/ga-dc/garnet/issues/524

PR #615, opened 2016-06-08
Show number of absences

tylercrosse:
This effectively undoes pull #530 based on discussion in issue #524. I think this is much more useful as a metric to manage students but maybe that's just me.


nolds9:
👍

nolds9:
I think number of absences is more helpful for both students and instructors

PR #627, opened 2016-07-28
Easier observations

jshawl:


app/assets/javascripts/nav.js
```diff
@@ -46,24 +46,6 @@ $(function() {
     timer = window.setTimeout(scrollNav(), 15); //delay of 15 ms
   });
 
-  // Smooth scroll
-  $('a').click(function(event){
```
jshawl: this was too much of a pita tbh

app/models/membership.rb
```diff
@@ -1,5 +1,7 @@
 class Membership < ActiveRecord::Base
   include ModelHelpers
+  include ActionView::Helpers::DateHelper
```
jshawl: this can be removed

nolds9:
@jshawl does this commit fix deep linking to obz? Currently unable to follow link to a specific observation - i.e. https://garnet.wdidc.org/memberships/825#observations-1448

PR #630, opened 2016-08-23
Fixes #608

### LINKED ISSUE #608: When manually creating an attendance, student has no status at all and therefore cannot check in

andrewsunglaekim:
Not sure whether to use string or this funky way to access `"unmarked"` in the test


andrewsunglaekim:
removed unmarked scope from mass creation of attendances on event.

PR #637, opened 2016-11-14
636/bugfix

andrewsunglaekim:
@nolds9 take a look at test and adjustment i made. @jshawl @amaseda additional :eyes: would be dope for 👍

### DIFF
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -68,8 +68,6 @@ class Cohort < ActiveRecord::Base
   def generate_events start_time, time_zone
     end_time = end_date.in_time_zone
     current_time = start_date.in_time_zone.change(hour: start_time)
-    diff = end_time - current_time
-    adjusted_end_time = current_time.since(diff)
     days_in_cohort = (end_date - start_date).to_i
 
     days_in_cohort.times do |i|
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -66,23 +66,19 @@ class Cohort < ActiveRecord::Base
   end
 
   def generate_events start_time, time_zone
-    # grabs offset leveraging current time of timezone passed in
-    # have to offset to accomodate for UTC, config timezone doesn't set "server time", rails always stores dates in UTC in the database
-    offset = ActiveSupport::TimeZone[time_zone].now.to_s.split(' ').last
+    end_time = end_date.in_time_zone
+    current_time = start_date.in_time_zone.change(hour: start_time)
+    diff = end_time - current_time
+    adjusted_end_time = current_time.since(diff)
     days_in_cohort = (end_date - start_date).to_i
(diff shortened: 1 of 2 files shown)


app/models/cohort.rb
```diff
@@ -68,8 +68,6 @@ def self.to_csv(memberships)
   def generate_events start_time, time_zone
```
nolds9: I don't think `time_zone` is used as input anymore for this method, we can remove this and the field on the form in the view

app/models/cohort.rb
```diff
@@ -66,23 +66,17 @@ def self.to_csv(memberships)
...
 
   def generate_events start_time, time_zone
-    # grabs offset leveraging current time of timezone passed in
-    # have to offset to accomodate for UTC, config timezone doesn't set "server time", rails always stores dates in UTC in the database
-    offset = ActiveSupport::TimeZone[time_zone].now.to_s.split(' ').last
+    end_time = end_date.in_time_zone
+    current_time = start_date.in_time_zone.change(hour: start_time)
     days_in_cohort = (end_date - start_date).to_i
-    current_day_time = start_date.to_datetime.change({hour: start_time, offset: offset })
-    # TODO: refactor using business_time gem to stream line all of this using matt's psuedocode
```
jshawl: can we keep TODO comment

app/models/cohort.rb
```diff
@@ -66,23 +66,17 @@ def self.to_csv(memberships)
...
     days_in_cohort.times do |i|
-      current_day_time += 1.day
+      current_time += 1.day
       # if current day is a weekday .wday will return a number between 1-5
-      if current_day_time.wday < 6 && current_day_time.wday > 0
+      if current_time.wday < 6 && current_time.wday > 0
         # sees if there's an event that has the same day and month as the current day
-        if !self.events.any?{|event| event.occurs_at.to_date == current_day_time.to_date}
-          self.events.create(occurs_at: current_day_time, title: current_day_time.strftime("%B %d, %Y"))
+        if !self.events.any?{|event| event.occurs_at.to_date == current_time.to_date}
```
jshawl: `unless` will make this block easier to read

spec/models/cohort_spec.rb
```diff
@@ -40,6 +47,11 @@
       num_events_after_running_method_again = test_cohort.events.length
       expect(num_events_before_running_method_again).to eq(num_events_after_running_method_again)
     end
+
+    it "accounts for DST" do
+      dst_test_cohort.generate_events(start_time, time_zone)
+      expect(dst_test_cohort.events.last.occurs_at.hour).to eq(dst_test_cohort.events.first.occurs_at.hour)
```
jshawl: 👍 radical

jshawl: can we add one more test for when the time springs forward

